PR #10, opened 2019-07-09
Add trial metadata model and db migration

jacoblurye:
Addresses [#506](https://trello.com/c/GiVijxC2/506-database-schema-updates-to-support-metadata-ingestion)

Adds the `TrialMetadata` model to the `models.py` module and registers it with the API. Also, includes an auto-generated database migration.


cidc-api/migrations/versions/37888cac6b2b_.py
```diff
@@ -0,0 +1,40 @@
+"""empty message
```
jacoblurye: This file was auto-generated by running `FLASK_APP=app.py flask db migrate` in the `cidc-api` directory.

cidc-api/app.py
```diff
@@ -23,17 +23,16 @@
...
-    with app.app_context():
-        app.logger.info("Upgrading the database...")
-        upgrade(MIGRATIONS)
-        app.logger.info("Done upgrading the database.")
+# Configure flask-migrate and upgrade the database
+# Note: while upgrades are performed automatically,
+# generating the migrations should be performed by hand
+# using the flask-migrate CLI, and the resulting files
+# should be checked into source control.
+Migrate(app, db)
```
jacoblurye: Now that we use a test database, we always want to run migrations on startup.

PR #16, opened 2019-07-15
Signed URL endpoint

jacoblurye:
Adds the endpoint `/ingestion/signed-upload-urls` that, given a request containing JSON specifying an upload "directory" and file names, returns a mapping of file names to signed upload URLs. This will allow a client (e.g., the cidc-cli) to upload directly to GCS using these URLs.

### DIFF
--- a/cidc-api/services/ingestion.py
+++ b/cidc-api/services/ingestion.py
@@ -2,13 +2,17 @@
 Endpoints for validating and ingesting metadata and data.
 """
 import json
+import datetime
 from typing import BinaryIO, Tuple
 
 from werkzeug.exceptions import BadRequest, InternalServerError, NotImplemented
 
+from google.cloud import storage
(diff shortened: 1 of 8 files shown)


cidc-api/services/ingestion.py
```diff
@@ -82,4 +86,70 @@ def validate():
...
+        "object_names": ["my-fastq-1.fastq.gz", "my-fastq-2.fastq.gz"]
+    }
+
+    Sample response body:
+    {
+        "my-fastq-1.fastq.gz": [a signed URL with PUT permissions],
+        "my-fastq-2.fastq.gz": [a signed URL with PUT permissions]
+    }
+    """
+    # Validate the request body
```
jacoblurye: Maybe going forward we should use JSON Schema or Cerberus for validations like this?

PR #37, opened 2019-08-09
User registration and initial RBAC

jacoblurye:
A first pass at user registration and role-based access control:
* Adds registration / approval related fields to the `Users` model
* Adds a `Permissions` model for representing access to trials / assays
* Adds `BearerAuth.role_auth`, which implements authorization logic based on a user's registration approval status and role
* Adds a pre-POST hook on the `users` resource enforcing that a requesting user can only create themself.


cidc_api/auth.py
```diff
@@ -38,11 +38,42 @@ def check_auth(
         TODO: role-based resource/method-level authorization
         """
         profile = self.token_auth(id_token)
+        is_authorized = self.role_auth(profile, allowed_roles, resource, method)
 
-        user = Users.create(profile["email"])
+        return is_authorized
 
+    def role_auth(
+        self, profile: dict, allowed_roles: List[str], resource: str, method: str
```
curlup: shouldn't roles be like enum or something?

cidc_api/models.py
```diff
@@ -69,18 +80,40 @@ class Users(CommonColumns):
     first_n = Column(String)
     last_n = Column(String)
     organization = Column(Enum(*ORGS, name="orgs"))
+    registration_approved = Column(
+        Boolean, nullable=False, default=False, server_default="false"
+    )
+    registration_approval_date = Column(DateTime)
```
curlup: wouldn't just one column `registration_approval_date` be enough to check for approval?

cidc_api/models.py
```diff
@@ -89,6 +122,28 @@ def create(email: str, session: Session = None):
...
+    granted_to_user = Column(
+        Integer, ForeignKey("users.id", ondelete="CASCADE"), index=True
+    )
+    trial_id = Column(
+        String,
+        ForeignKey("trial_metadata.trial_id", ondelete="CASCADE"),
+        nullable=False,
+        index=True,
+    )
+    assay_type = Column(Enum(*SUPPORTED_ASSAYS, name="assays"), nullable=False)
```
curlup: I just wonder - there was such a requirement that perms are given wrt to specific assays only? 
If not - wouldn't it be a permission granted always on all assays, so we will end up storing tens of perm entries instead of just one?

cidc_api/auth.py
jacoblurye: It's true that it would be accurate to model roles with a type like `Role = Union[Literal['cidc-admin'], Literal['cimac-user'], ...]`, meaning that `allowed_roles: List[Role]` in the `role_auth` function signature, but I'm not sure we get anything in practice from adding this complexity to the type signature - especially since the `BearerAuth` class extends `eve.auth.TokenAuth`, whose function signature only guarantees that `check_auth` (and subsequently `role_auth`) will be passed `allowed_roles: List[str]`.

cidc_api/models.py
jacoblurye: Definitely. I was replicating the user model from the old API [here](https://github.com/CIMAC-CIDC/cidc-api/blob/1877907119efd30c7698ecc7c5da84f877ba02d7/schemas/user_schema.py#L33), which has both `approved` and `registration_approval_date`. I'll make this simplification, though.

cidc_api/models.py
jacoblurye: I don't know of an explicit specification, so this was a bit speculative. Maybe making the `assay` field nullable would allow us to represent both trial-wide and trial-assay-level permissions (i.e., if empty, then the user can view all assays for that trial)? Alternatively, `assay_type` could instead be an array `assay_types`, containing all the assays a user is permitted to view.

jacoblurye:
About your roles question, we should double check with @priti88 and @jim-bo, but this list is copied from the "User Roles and Access Privileges" section of the SSP.

cidc_api/auth.py
curlup: I'm not sure I was able to follow your line of thought, I guess a list of strings will do.

cidc_api/models.py
curlup: I see. I think it will probably be okay to keep it like it is, if we don't expect a sheer number of users. And from my understanding we don't.

PR #42, opened 2019-08-14
Downloadable files

jacoblurye:
A starting point for the `DownloadableFiles` model that stores required [`artifact_core`](https://github.com/CIMAC-CIDC/cidc-schemas/blob/master/cidc_schemas/schemas/artifacts/artifact_core.json) properties for a file. We will likely add more non-artifact metadata to this model going forward, like participant/sample IDs.


cidc_api/models.py
```diff
@@ -20,10 +22,70 @@
...
+
+# See: https://github.com/CIMAC-CIDC/cidc-schemas/blob/master/cidc_schemas/schemas/artifacts/artifact_core.json
+ARTIFACT_CATEGORIES = [
+    "Assay Artifact from CIMAC",
+    "Pipeline Artifact",
+    "Manifest File",
+    "Other",
+]
+ASSAY_CATEGORIES = [
+    "Whole Exome Sequencing (WES)",
```
curlup: For the sake of consistency it would be nice to have `ASSAY_CATEGORIES`, `ARTIFACT_CATEGORIES` to be parsed from the schema directly and not repeated here.

E.g. in `dm_update` schema branch I already removed "Other" from both of those.

jacoblurye: Yeah, consistency will be a problem here. I worry, though, that populating these constants dynamically from `cidc_schemas` may lead to inconsistencies between the database schema and the python data models. Since the database migrations need to be generated by hand, a dynamic change to `ASSAY_CATEGORIES` won't automatically be reflected in the Postgres schema.

Alternatively, we could just make `DownloadableFiles.assay_category` and  `DownloadableFiles.artifact_category` string-type rather than enum type, relying on `cidc_schemas` to do the validation for us.

PR #46, opened 2019-08-19
Add signed GCS download URL to downloadable_files

jacoblurye:
This PR creates an Eve event hook that adds GCS signed URLs to fetched downloadable files. This should make file downloads from the portal UI work without any changes to the UI code.


cidc_api/services/files.py
```diff
@@ -0,0 +1,36 @@
...
+
+def insert_download_urls(payload: dict):
+    """
+    Get a signed GCS download URL for each of the requested files.
+    """
+    # Each call to insert_download_url generates a request
+    # to the GCS API, so we get a speed-up from multithreading:
+    # although Python dispatches the requests synchronously,
+    # the OS handles the network requests in parallel.
+    with ThreadPool() as pool:
```
curlup: Why is this ThreadPool needed?

jacoblurye: Because every invocation to `insert_download_url` makes a request to the Google Cloud Storage API. If there are a lot of files in `payload`, handling each of these requests synchronously would take a very long time. By dispatching these requests to a `ThreadPool`, the network operations will be made in parallel.

PR #57, opened 2019-08-29
Update README.md wrt to pytest GOOGLE_APPLICATION_CREDENTIALS

curlup:


README.md
```diff
@@ -119,6 +119,11 @@ To run the tests, simply run:
 pytest
 ```
 
+You might first need to set up something like
```
jacoblurye: I might say something more general, like

> To run the tests, you need to set the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to a path to JSON file containing your gcloud credentials. To create this file if it doesn't exist already, run:
> ```
> gcloud auth application-default login
> ```
> Set `GOOGLE_APPLICATION_CREDENTIALS` to the path listed after `"Credentials saved to file: "` in the output of the above command.

curlup: How do we make sure that `prod` env is not touched by accident?

curlup:
@jacoblurye do we still need this or tests now ignore those creds and we can just close this PR?

PR #62, opened 2019-09-04
/upload_assay generates upload uris w/o local_path, only with timestamp

curlup:
and models know how to strip this timestamp off and how to prism.merge_artifact.


tests/services/test_ingestion.py
```diff
@@ -167,7 +167,9 @@ def test_upload_wes(
     gcs_object_name = url_mapping[local_path]
     assert local_path in url_mapping
     assert gcs_object_name.startswith(gcs_prefix)
-    assert gcs_object_name.endswith(local_path)
+    assert not gcs_object_name.endswith(
```
jacoblurye: Should this check `local_path in gcs_object_name` rather than just `endswith`?

PR #71, opened 2019-09-06
Use version range for cidc-schemas

jacoblurye:
The intention here is to automatically upgrade `cidc-schemas` to any newer version without breaking changes to its API.


curlup:
What is this new tests/services/data/.DS_Store file? @jacoblurye

PR #78, opened 2019-09-09
ingestion.upload_manifest

curlup:


cidc_api/services/ingestion.py
```diff
@@ -136,7 +136,22 @@ def upload_manifest():
     """
     schema_hint, schema_path, xlsx_file = extract_schema_and_xlsx()
 
-    return NotImplemented("Manifest ingestion is not yet supported.")
+    md_patch, file_infos = prism.prismify(xlsx_file, schema_path, schema_hint)
+
+    if len(file_infos) > 0:
+        raise BadRequest(f"Unexpected local files in template for {schema_hint}")
```
jacoblurye: Maybe "Shipping manifests shouldn't reference any local files"?

cidc_api/services/ingestion.py
```diff
@@ -136,7 +136,22 @@ def upload_manifest():
...
+
+    try:
+        trial_id = md_patch[TRIAL_ID_FIELD]
+    except KeyError:
+        raise BadRequest(f"No {TRIAL_ID_FIELD} found/parsed from template.")
+
+    try:
+        TrialMetadata.patch_manifest(trial_id, md_patch)
+    except AssertionError as e:
+        raise BadRequest(f"Trial with {TRIAL_ID_FIELD} {trial_id} not found.")
```
jacoblurye: Can we be totally certain nothing else in `TrialMetadata.patch_manifest` will throw an assertion error?

cidc_api/services/ingestion.py
```diff
@@ -136,7 +136,22 @@ def upload_manifest():
...
+
+    return jsonify({})
```
jacoblurye: Maybe we can return the manifest patch here - might be useful to display some components of this as feedback in the UI 🤷‍♂

tests/services/test_ingestion.py
```diff
@@ -118,14 +126,30 @@ def test_extract_schema_and_xlsx_failures(app, url, data, error, message):
...
+    app_no_auth, pbmc_valid_xlsx, test_user, db_with_trial_and_user
+):
     """Ensure the upload_manifest endpoint follows the expected execution flow"""
     client = app_no_auth.test_client()
 
     res = client.post(
         MANIFEST_UPLOAD, data=form_data("pbmc.xlsx", pbmc_valid_xlsx, "pbmc")
     )
-    assert res.status_code == 501  # Not Implemented
+    res_body_json = res.json
```
jacoblurye: Is `res_body_json` used? Also, `# Not yet Implemented` no longer applies, right?

tests/test_models.py
```diff
@@ -91,8 +91,10 @@ def test_trial_metadata_patch_manifest(db):
 
     # Look the trial up and check that it has the participant in it
     trial = TrialMetadata.find_by_trial_id(TRIAL_ID)
+    trial_metadata_json_recheck = trial.metadata_json
```
jacoblurye: what's the purpose of this change?

cidc_api/models.py
```diff
@@ -268,32 +284,30 @@ def patch_manifest(trial_id: str, manifest_patch: dict, session: Session):
 
             TODO: apply this update directly to the not-yet-existent TrialMetadata.assays field
         """
-        TrialMetadata.patch_trial_metadata(trial_id, manifest_patch, session=session)
+        trial = TrialMetadata.select_for_update_by_trial_id(trial_id)
```
jacoblurye: since this code is repeated verbatim above, maybe we should make a function that wraps this:
```python
@staticmethod
@with_default_session
def _select_and_patch_trial_metadata(trial_id: str, json_patch: dict, session: Session):
    trial = TrialMetadata.select_for_update_by_trial_id(trial_id)
    trial._patch_trial_metadata(assay_patch, session=session)
```
We still get the benefits of row-level locking on `trial` (as I understand it).

cidc_api/gcloud_client.py
```diff
@@ -33,21 +33,33 @@ def _iam_id(user_email: str) -> str:
...
-    blob = bucket.blob(blob_name)
+    upload_bucket: storage.Bucket = _get_bucket(GOOGLE_UPLOAD_BUCKET)
+    blob = upload_bucket.blob(blob_name)
     blob.upload_from_file(filebytes)
-    return blob_name
+
+    data_bucket = _get_bucket(GOOGLE_DATA_BUCKET)
+    final_object = upload_bucket.copy_blob(blob, data_bucket)
+
+    return blob
```
jacoblurye: Should we return `final_object` here instead? Only might make a difference for the `time_created` attribute.

tests/test_gcloud_client.py
```diff
@@ -46,11 +47,17 @@ def set_iam_policy(self, policy):
...
 
-    uri = upload_xlsx_to_gcs(trial, template_type, assay_type, BytesIO(b"12345"))
+    uri = upload_xlsx_to_gcs(
+        trial,
+        template_type,
+        assay_type,
+        BytesIO(b"12345"),
+        datetime.datetime.now().isoformat(),
+    )
     assert trial in uri
```
jacoblurye: This and the below assertions will need to be updated now that `upload_xlsx_to_gcs` returns a `Blob`, not a URI.

PR #84, opened 2019-09-13
Trial-Assay Permissions and Role-Based Access Control

jacoblurye:
This PR introduces trial-assay level permissions and more stringent RBAC into the API.

~Note: this PR should probably not be merged until there is a way to award users new trial-assay permissions through the UI.~ (actually, having an admin interface for this isn't critical, so we should merge when the PR is ready and plan on adding user permissions through postgres)

**Trial-Assay Level Permissions**: When a user GETs `downloadable_files`, the `services.files.update_file_filters` hook adds filters to their query based on the contents of the `permissions` table so that the results of the GET request only include files the user is allowed to see.

**RBAC**: Check out `models.get_DOMAIN` for an overview of what constraints are enforced (interested in feedback on whether these appear to be the right constraints among other things. Also, note that RBAC constraints are at the API-level, not the database level, so anyone with database access can currently perform any operation permitted by the database schema.


cidc_api/services/files.py
```diff
@@ -1,15 +1,70 @@
...
+        request.args = ImmutableMultiDict(lookup)
+        return
+
+    # Build a where-query that looks up only the downloadable_files with
+    # trial ID and assay type that the current user is allowed to view.
+    # If the user has permission to view WES for Trial "1" and Olink for "2",
+    # this query will look like:
+    #   (trial==1 and assay_type==wes)or(trial==2 and assay_type==olink)
+    where_query = "or".join(
+        f'(trial=="{p.trial_id}" and assay_type=="{p.assay_type}")' for p in permissions
```
curlup: There's a possibility for query injection here
like 
```
> print('THIS="{that}"'.format(that='some" or THIS="anything'))
THIS="some" or THIS="anything"
```
 
```suggestion
        f'(trial=={p.trial_id!r} and assay_type=={p.assay_type!r})' for p in permissions
```

cidc_api/services/files.py
```diff
@@ -1,15 +1,70 @@
...
+                "Mongo-style JSON filters are not supported on the downloadable_files table. "
+                "Use SQL-style filters with the form "
+                "'<field1> == <value1> or (<field1> == <value2> and <field3> = <value3>)"
+            )
+        except json.JSONDecodeError:
+            # The user provided a non-JSON filter, which is what we want
+            pass
+
+        # Add permission filters to the user's where query in an "and" clause
+        where_query = f"({user_where_query})and({where_query})"
```
curlup: here too
```
> print("({0})and({1})".format('A = "B") or (A = "B"', 'c="d"'))
(A = "B") or (A = "B")and(c="d")
```

PR #85, opened 2019-09-13
only supported assays in info

curlup:


cidc_api/services/info.py
```diff
@@ -2,11 +2,21 @@
...
 
 from cidc_schemas.template import _TEMPLATE_PATH_MAP
+from cidc_schemas import prism
+
 
 info_api = Blueprint("info", __name__, url_prefix="/info")
 
-assay_list = [k for k, v in _TEMPLATE_PATH_MAP.items() if "metadata" in v]
-manifest_list = [k for k, v in _TEMPLATE_PATH_MAP.items() if "manifests" in v]
+assay_list = [
```
jacoblurye: This should be equivalent to `assay_list = prism.SUPPORTED_ASSAYS`

cidc_api/services/info.py
```diff
@@ -2,11 +2,21 @@
...
+    k
+    for k, v in _TEMPLATE_PATH_MAP.items()
+    if "metadata" in v and k in prism.SUPPORTED_ASSAYS
+]
+manifest_list = [
```
jacoblurye: `manifest_list = prism.SUPPORTED_MANIFESTS`

PR #88, opened 2019-09-16
Olink to Cimac_ID linkage

roshni-b:
WIP


cidc_api/services/ingestion.py
```diff
@@ -329,3 +356,14 @@ def signed_upload_urls():
         object_urls[object_name] = object_url
 
     return jsonify(object_urls)
+
+@ingestion_api.route("/extra-metadata", methods=["POST"])
```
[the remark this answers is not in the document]
roshni-b: renamed to `/extra-assay-metadata`

cidc_api/services/ingestion.py
```diff
@@ -301,6 +309,9 @@ def upload_assay():
         "url_mapping": url_mapping,
         "gcs_bucket": GOOGLE_UPLOAD_BUCKET,
     }
+    if schema_hint in prism.ASSAYS_WITH_EXTRA_METADATA:
```
curlup: shouldn't this just rely on `files_with_extra_md` list being not empty? then we won't need this extra prism dependency (tighter coupling).

cidc_api/models.py
```diff
@@ -561,6 +561,15 @@ def create(
 
         return job
 
+    @staticmethod
+    @with_default_session
+    def merge_extra_metadata(job_id, artifact_uuid, file, session):
+
+        job = AssayUploads.find_by_id(job_id, session = session)
+        prism.merge_artifact_extra_metadata(job.assay_patch, artifact_uuid, job.assay_type, file)
```
curlup: so how does prism call result gets saved? am I missing something or is it ignored here? So we need to save it?

Also I would make it a bulk method accepting multiple uuids/files in one, so we won't have to commit multiple times (which seems wrong)

cidc_api/services/ingestion.py
```diff
@@ -440,3 +451,44 @@ def signed_upload_urls():
...
+        raise BadRequest("Expected job_id in form")
+
+    if not request.files:
+        raise BadRequest("Expected files in request (mapping from artifact uuids to open files)")
+
+    job_id = request.form['job_id']
+
+    for uuid, file in request.files.to_dict().items():
+        try:
+            AssayUploads.merge_extra_metadata(job_id, uuid, file)
```
curlup: I think would be better to merge them all in one, avoiding committing session multiple times.

tests/services/test_ingestion.py
```diff
@@ -493,3 +496,39 @@ def test_poll_upload_merge_status(app, db, test_user, monkeypatch):
...
+    form = {
+        'job_id': 123,
+        'uuid-1': (io.BytesIO(b'fake file 1'), 'fname1'),
+        'uuid-2': (io.BytesIO(b'fake file 2'), 'fname2')
+    }
+
+    res = client.post('/ingestion/extra-assay-metadata', data=form)
+
+    assert res.status_code == 200
+    merge_extra_metadata.assert_called()
```
curlup: please add assertions on data being actually updated/saved.

curlup: also check validation errors being handled and maybe converted to bad request errors with human comprehensible message.

roshni-b: BadRequest errors being raised [here](https://github.com/CIMAC-CIDC/cidc-api-gae/blob/e4153aca6619be85e01c95fcd48d1fbb8a7f6ab4/cidc_api/services/ingestion.py#L524)

roshni-b: the prism call result updates the existing patch (referred [here](https://github.com/CIMAC-CIDC/cidc-schemas/blob/f8dfa6c63076a95d6f8e155f89998431981d4caa/cidc_schemas/prism.py#L786))

cidc_api/services/ingestion.py
```diff
@@ -342,7 +342,7 @@ def upload_assay(
         "url_mapping": url_mapping,
         "gcs_bucket": GOOGLE_UPLOAD_BUCKET,
     }
-    if schema_hint in prism.ASSAYS_WITH_EXTRA_METADATA:
+    if bool(files_with_extra_md):
```
curlup: no need to explicitly convert to bool =)

PR #91, opened 2019-09-16
Add `/users/self` endpoint and allow users to read own permissions

jacoblurye:
Since only `cidc-admins` can access the `users` resource according to current RBAC configs, add a `/users/self` endpoint that allows an authenticated user with any role to access their own profile information. Also, remove the now-defunct `filter_user_lookups` hook.

Also, adds `services.permissions.update_permissions_filters` to allow non-admin users to read their own permissions.

### DIFF
--- a/cidc_api/models.py
+++ b/cidc_api/models.py
@@ -111,7 +111,7 @@ def get_DOMAIN() -> dict:
 
     # Restrict operations on the 'permissions' resource:
     # * Only admins can write or update 'permissions'
-    # * All users can read permissions, but the results will be filtered by
+    # * All users can list the 'permissions' resource, but the results will be filtered by
     #   services.permissions.update_permission_filters to only include their permissions
     domain["permissions"]["allowed_write_roles"] = [CIDCRole.ADMIN.value]
     domain["permissions"]["allowed_item_roles"] = [CIDCRole.ADMIN.value]
--- a/cidc_api/models.py
+++ b/cidc_api/models.py
@@ -105,11 +105,16 @@ def get_DOMAIN() -> dict:
     domain["new_users"]["resource_methods"] = ["POST"]
 
     # Restrict operations on resources that only admins should be able to access
-    for resource in ["users", "permissions", "trial_metadata"]:
+    for resource in ["users", "trial_metadata"]:
         domain[resource]["allowed_roles"] = [CIDCRole.ADMIN.value]
         domain[resource]["allowed_item_roles"] = [CIDCRole.ADMIN.value]
 
-    # Make permissions deletable by admins
+    # Restrict operations on the 'permissions' resource:
+    # * Only admins can write or update 'permissions'
+    # * All users can read permissions, but the results will be filtered by
+    #   services.permissions.update_permission_filters to only include their permissions
+    domain["permissions"]["allowed_write_roles"] = [CIDCRole.ADMIN.value]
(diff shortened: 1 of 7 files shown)


cidc_api/services/permissions.py
```diff
@@ -0,0 +1,25 @@
...
+    the current user's permissions, unless the user is an admin.
+    """
+    user = _request_ctx_stack.top.current_user
+
+    # Admins can get all permissions
+    if user.role == CIDCRole.ADMIN.value:
+        return
+
+    # Otherwise, only include permissions granted to this user
+    lookup["to_user"] = user.id
```
curlup: I though you said that such `lookup` don't get populated and processed for eve-sqlalchemy?

cidc_api/models.py
```diff
@@ -105,11 +105,16 @@ def get_DOMAIN() -> dict:
...
         domain[resource]["allowed_roles"] = [CIDCRole.ADMIN.value]
         domain[resource]["allowed_item_roles"] = [CIDCRole.ADMIN.value]
 
-    # Make permissions deletable by admins
+    # Restrict operations on the 'permissions' resource:
+    # * Only admins can write or update 'permissions'
+    # * All users can read permissions, but the results will be filtered by
+    #   services.permissions.update_permission_filters to only include their permissions
+    domain["permissions"]["allowed_write_roles"] = [CIDCRole.ADMIN.value]
+    domain["permissions"]["allowed_item_roles"] = [CIDCRole.ADMIN.value]
```
curlup: I see
>* All users can read permissions, but the results will be filtered by

but then only
> domain["permissions"]["allowed_item_roles"] = [CIDCRole.ADMIN.value]

Am I missing something ?

cidc_api/services/users.py
```diff
@@ -18,14 +18,25 @@
 from models import Users, CIDCRole
 from emails import new_user_registration, confirm_account_approval
 
+users_api = Blueprint("users", __name__, url_prefix="/users")
+
+
+@users_api.route("/self", methods=["GET"])
+@requires_auth("users.self")
+def get_self():
+    jsonish_user = _request_ctx_stack.top.current_user.__dict__
```
curlup: Hm, seems kind'a hacky =}
What if sql-alchemy or something will start adding some other field? 
Shouldn't there be like a white-list based on the `models.User._fields` or something like that?

cidc_api/services/permissions.py
jacoblurye: It doesn't get populated (`lookup` is always empty as an argument to this function), but it does get processed (or it least the passing tests suggest this is the case 🤷‍♂)

cidc_api/models.py
jacoblurye: Eve distinguishes between resource level requests to, e.g., `/permissions`, and item-level requests to `/permissions/<item id>`. There isn't a reason to give non-admins item-level access, so that's line 117. I can update the comment, though.

cidc_api/services/users.py
jacoblurye: Oh, yeah - `Permissions.__table__.columns.keys()` would give the list of valid columns. I'll update to use that.

cidc_api/models.py
jacoblurye: ^this is why a non-admin user gets 200 from a GET to `/permissions` in test_auth.py

jacoblurye:
lol

PR #94, opened 2019-09-18
Add @resource decorator and fix users/self rbac handling

jacoblurye:


cidc_api/services/utils.py
```diff
@@ -0,0 +1,43 @@
...
+            kwargs["resource"] = name
+            app.config["DOMAIN"][name] = {
+                "public_methods": public_methods,
+                "allowed_roles": allowed_roles,
+                "allowed_read_roles": allowed_read_roles,
+                "allowed_write_roles": allowed_write_roles,
+                "authentication": app.auth,
+            }
+
+            @requires_auth("resource")
```
curlup: shouldn't this be like 
```suggestion
            @requires_auth(name)
```

jacoblurye: Eve's `requires_auth` decorator expects an "endpoint class" (either `"resource"`, `"item"`, or `"home"`) as its argument: https://github.com/pyeve/eve/blob/4b052b843f7eb1b402709d8319332a818b35c52d/eve/auth.py#L16

PR #96, opened 2019-09-23
Different custom rbac approach

jacoblurye:
Reverts https://github.com/CIMAC-CIDC/cidc-api-gae/pull/94, which is leading to `KeyError: 'internal_resource'` getting thrown on GET requests to `/` in the API. Turns out those changes run into the issues documented here: https://github.com/pyeve/eve/issues/860

This solution is simpler, implementing a custom `requires_auth` decorator that behaves the way I _thought_ `eve.auth.requires_auth` was supposed to.


cidc_api/services/ingestion.py
```diff
@@ -201,7 +209,9 @@ def upload_manifest():
 
 
 @ingestion_api.route("/upload_assay", methods=["POST"])
-@resource("ingestion/upload_assay")
+@requires_auth(
+    "ingestion/upload_manifest", [CIDCRole.ADMIN.value, CIDCRole.CIMAC_BIOFX_USER.value]
```
curlup:
```suggestion
    "ingestion/upload_assay", [CIDCRole.ADMIN.value, CIDCRole.CIMAC_BIOFX_USER.value]
```

cidc_api/services/users.py
```diff
@@ -16,14 +16,13 @@
...
-
-from .utils import resource
+from auth import requires_auth
 
 users_api = Blueprint("users", __name__, url_prefix="/users")
 
 
 @users_api.route("/self", methods=["GET"])
-@resource("users/self")
+@requires_auth("self")
```
curlup: AFAIK this will be transformed to 
`app.auth.authorized(allowed_roles=[], ...)`, will this allow everybody then or something? If so, can you might be add an explicit comment to the decorator doc.

cidc_api/services/ingestion.py
```diff
@@ -84,7 +83,14 @@ def extract_schema_and_xlsx() -> Tuple[str, str, BinaryIO]:
 
 
 @ingestion_api.route("/validate", methods=["POST"])
-@resource("ingestion/validate")
+@requires_auth(
+    "ingestion/validate",
```
curlup: if it's actually just for manifest validation, should it disallow `CIMAC_BIOFX_USER` then?

PR #99, opened 2019-09-25
Publish to the patient_sample_update topic on manifest ingestion

jacoblurye:
When a trial's patient/sample metadata is updated via a manifest upload, the API will publish that trial's ID to the `patient_sample_update` Pub/Sub topic.


jacoblurye:
(see: https://github.com/CIMAC-CIDC/cidc-devops/pull/26)

PR #111, opened 2019-10-02
downloadable_files dedup

curlup:


migrations/versions/aa8a94b8e115_.py
```diff
@@ -0,0 +1,64 @@
...
+down_revision = "08591dac6d5c"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.execute(
+        """
+        DELETE from
```
jacoblurye: Fancy query you got here 🙂 Maybe add a comment explaining that it will delete all but the most recently-updated downloadable file for each `object_url`?

tests/services/test_ingestion.py
```diff
@@ -202,6 +202,36 @@ def test_upload_manifest(
     mocks.make_all_assertions()
 
 
+def test_upload_manifest_twice(
+    app_no_auth, some_file, test_user, db_with_trial_and_user, db, monkeypatch
+):
+    """Ensure the upload_manifest endpoint follows the expected execution flow"""
```
jacoblurye: Could use a more specific docstring

PR #113, opened 2019-10-03
file dedup in prod

curlup:


jacoblurye:
Can we wait till #114 is integrated?

curlup:
waiting for admin manifest upload perms fix by @jacoblurye

PR #119, opened 2019-10-07
Artifact specific metadata

jacoblurye:
The next step will be to update the `ingest_upload` cloud function to create assay downloadable files with `additional_metadata` populated.


cidc_api/models.py
```diff
@@ -636,6 +637,7 @@ def create_from_metadata(
         assay_type: str,
         file_metadata: dict,
         session: Session,
+        additional_metadata: dict = {},
```
curlup: You don't want to do mutable values as defaults in python =) 
https://lmgtfy.com/?q=python+argument+default+dict

PR #127, opened 2019-10-10
672 manifest full validation

curlup:
Will address in future: 
* prismify/merge errors should have context in terms of Excel rows/cells.
* add saving "manifest draft" to the DB on validation, and then only apply it on submission.

### DIFF
--- a/cidc_api/services/ingestion.py
+++ b/cidc_api/services/ingestion.py
@@ -58,7 +58,7 @@ def extract_schema_and_xlsx() -> Tuple[str, str, BinaryIO]:
         BadRequest: if the above requirements aren't satisfied
 
     Returns:
-        Tuple[str, str, dict]: the schema identifier (aka template type), the schema path, and the open xlsx file
+        Tuple[Template, BinaryIO]: template, and the open xlsx file
     """
     # If there is no form attribute on the request object,
     # then either one wasn't supplied, or it was malformed
@@ -82,7 +82,12 @@ def extract_schema_and_xlsx() -> Tuple[str, str, BinaryIO]:
         raise BadRequest("Expected a form entry for 'schema'")
     schema_id = schema_id.lower()
 
-    return schema_id, xlsx_file
+    try:
+        template = Template.from_type(schema_id)
+    except Exception:
+        raise BadRequest(f"Unknown template type {schema_id!r}")
+    
+    return template, xlsx_file
 
 
 def validate(template, xlsx):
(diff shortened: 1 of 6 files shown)


cidc_api/services/ingestion.py
```diff
@@ -199,18 +203,36 @@ def wrapped(*args, **kwargs):
...
+    user: Users,
+    trial: TrialMetadata,
+    template_type: str,
+    xlsx_file: BinaryIO,
+    md_patch: dict,
+    file_infos: List[prism.LocalFileUploadEntry],
+):
+    # Validation is done within `upload_handler`
+    # so we just return ok here
+    return jsonify({"errors": []})
```
jacoblurye: Shouldn't this return collected validation errors?

curlup: as this is wrapped with `upload_handler` that will return all validation errors, here we now validation has passed so no errors

jacoblurye: I'd make the signature `def validate_endpoint(*args, **kwargs)` to make it more explicit that none of those arguments are used.

cidc_api/services/ingestion.py
```diff
@@ -325,10 +347,8 @@ def upload_assay(
         if file_info.metadata_availability:
             files_with_extra_md[file_info.local_path] = file_info.upload_placeholder
 
-    # Upload the xlsx template file to GCS
-    xlsx_file.seek(0)
```
jacoblurye: Is it safe to remove this?

tests/services/test_ingestion.py
```diff
@@ -89,28 +89,32 @@ def form_data(filename=None, fp=None, schema=None):
...
     mocks = UploadMocks(monkeypatch)
-    mocks.validate_excel.return_value = ["error-1"]
+    mocks.iter_errors.return_value = ["test error"]
     client = app_no_auth.test_client()
     data = form_data("pbmc.xlsx", some_file, "pbmc")
     res = client.post(VALIDATE, data=data)
-    assert res.status_code == 200
-    assert len(res.json["errors"]) > 0
+    assert res.status_code == 400
+    assert len(res.json["_error"]["message"]) > 0
```
jacoblurye: We're going to need to handle this change in the portal (and, I think, the CLI).

curlup: I think it's handled already, that's why I felt like changing it to this should be okay.


https://github.com/CIMAC-CIDC/cidc-ui/blob/94c6391b8c17ace51f4e9634abc72400f5a02c95/src/api/api.ts#L34

https://github.com/CIMAC-CIDC/cidc-cli/blob/aec60df3c6f0a9361b41d74dfe4323f87b7a8b97/cli/api.py#L23

cidc_api/services/ingestion.py
curlup: I'll double check.

tests/services/test_ingestion.py
jacoblurye: So, it's handled in the sense that there won't be a runtime error, but if `message` is an array of values, it will print out as a JSON array in both cases, which isn't very user-friendly. As such, I think it's fine to merge as is, but we might want to make tickets about handling `message` arrays.

cidc_api/services/ingestion.py
curlup: moved within `upload_xlsx_to_gcs`

tests/services/test_ingestion.py
curlup: I checked UI at least there it handles array of error messages in a proper way. So I'm going with it.

PR #130, opened 2019-10-18
support for disabled users

curlup:


tests/test_auth.py
```diff
@@ -223,6 +223,22 @@ def test_role_auth(bearer_auth, app, db):
         # If the resource has no role restrictions, they should be authorized
         assert bearer_auth.role_auth(profile, [], "some-resource", "some-http-method")
 
+    # Disable user
+    db.query(Users).filter_by(email=EMAIL).update(dict(disabled=True))
+    db.commit()
+
+    with app.test_request_context():
```
jacoblurye: Should this also check that a disabled user can `GET` `/users/self`?

curlup:
> I figure there will need to be some accompanying UI changes? `user.disabled` isn't handled in the portal currently.

It will show unauth error as it is for manifest validation attempt - and that's what I wanted.

PR #131, opened 2019-10-18
Add download_url endpoint for explicitly getting signed URLs for downloadable files

jacoblurye:
Accompanying UI updates are needed.

### DIFF
--- a/.env
+++ b/.env
@@ -9,14 +9,14 @@ DEBUG=True
 TEST_POSTGRES_URI = 'postgresql://cidcdev:1234@localhost:5432/cidctest'
 
 # Connect to a local postgres instance
-POSTGRES_URI='postgresql://cidcdev:1234@localhost:5432/cidc'
+# POSTGRES_URI='postgresql://cidcdev:1234@localhost:5432/cidc'
 
 # Connect to our Cloud SQL Postgres instance.
 # Note: POSTGRES_URI must be commented out for this to work
-# CLOUD_SQL_SOCKET_DIR='<YOUR HOME DIRECTORY>/.cloudsql/'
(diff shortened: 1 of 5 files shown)


.env
```diff
@@ -9,14 +9,14 @@ DEBUG=True
...
-POSTGRES_URI='postgresql://cidcdev:1234@localhost:5432/cidc'
+# POSTGRES_URI='postgresql://cidcdev:1234@localhost:5432/cidc'
 
 # Connect to our Cloud SQL Postgres instance.
 # Note: POSTGRES_URI must be commented out for this to work
-# CLOUD_SQL_SOCKET_DIR='<YOUR HOME DIRECTORY>/.cloudsql/'
-# CLOUD_SQL_INSTANCE_NAME='cidc-dfci-staging:us-central1:cidc-postgresql'
-# CLOUD_SQL_DB_USER='cidcuser'
-# CLOUD_SQL_DB_NAME='cidc'
+CLOUD_SQL_SOCKET_DIR='/Users/jlurye/.cloudsql/'
```
curlup: hm

cidc_api/services/files.py
```diff
@@ -5,17 +5,53 @@
...
+
+    # Ensure user has permission to access this file
+    perms = Permissions.find_for_user(user)
+    if (
+        not any(
+            perm.assay_type == file_record.assay_type
+            and perm.trial_id == file_record.trial_id
+            for perm in perms
+        )
+        and user.role != CIDCRole.ADMIN.value
```
curlup: for the sake of readability I suggest rewrite this into nested `if`s form or something.

.env
jacoblurye: oops

PR #140, opened 2019-10-28
Bump cidc-schemas version

jacoblurye:
...to deploy changes in https://github.com/CIMAC-CIDC/cidc-schemas/pull/211


curlup:
I thought it should pull newest schemas even without this. Huh?

PR #142, opened 2019-10-28
Handle CLI version on authenticated endpoints

jacoblurye:
This PR sets up the API to reject requests on authenticated endpoints from clients who appear to be using outdated versions of the CIDC-CLI.

### DIFF
--- a/cidc_api/auth.py
+++ b/cidc_api/auth.py
@@ -82,7 +82,39 @@ class BearerAuth(TokenAuth):
 
         log_user_and_request_details(is_authorized)
 
+        self.enforce_cli_version()
+
         return is_authorized
+
+    def enforce_cli_version(self):
+        """
+        If the current request appears to come from the CLI and not the Portal, enforce the configured
+        minimum CLI version.
+        """
+        user_agent = request.headers.get("User-Agent")
+
+        # e.g., during testing no User-Agent header is supplied
+        if not user_agent:
+            return
+
+        client, version = user_agent.split("/")
+
+        # Old CLI versions don't update the User-Agent header, so we (perhaps dangerously)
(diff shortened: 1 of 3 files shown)


cidc_api/auth.py
```diff
@@ -82,8 +82,40 @@ def log_user_and_request_details(is_authorized: bool):
...
+        client, version = user_agent.split("/")
+
+        # Old CLI versions don't update the User-Agent header, so we (perhaps dangerously)
+        # assume any request coming from the python requests library is from a "very" old
+        # version of the CLI.
+        is_very_old_cli = client == "python-requests"
+
+        # Newer version of the CLI update the User-Agent header to `cidc-cli/{version}`,
+        # so we can assess whether the requester needs to update their CLI.
+        is_old_cli = client == "cidc-cli" and version < app.config["MIN_CLI_VERSION"]
```
curlup: I don't think that's a proper way to compare versions - just like strings

cidc_api/auth.py
```diff
@@ -82,8 +82,40 @@ def log_user_and_request_details(is_authorized: bool):
...
+        # e.g., during testing no User-Agent header is supplied
+        if not user_agent:
+            return
+
+        client, version = user_agent.split("/")
+
+        # Old CLI versions don't update the User-Agent header, so we (perhaps dangerously)
+        # assume any request coming from the python requests library is from a "very" old
+        # version of the CLI.
+        is_very_old_cli = client == "python-requests"
```
curlup: what if someone will use python-requests in their own client impl 😅

jacoblurye: Hence `perhaps dangerously` 🙂  

Happy to take this out, though.

PR #143, opened 2019-10-28
allows to call CFns by http in dev mode

curlup:
acc by https://github.com/CIMAC-CIDC/cidc-cloud-functions/pull/63

### DIFF
--- a/cidc_api/config/settings.py
+++ b/cidc_api/config/settings.py
@@ -19,7 +19,7 @@ TESTING = environ.get("TESTING") == "True"
 ## End application environment config
 
 ## Configure Dev CFn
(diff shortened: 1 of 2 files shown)


cidc_api/config/settings.py
```diff
@@ -18,6 +18,10 @@
 TESTING = environ.get("TESTING") == "True"
 ## End application environment config
 
+## Configure Dev CFn
+CFNS_HTTP = environ.get("CFNS_HTTP")
```
jacoblurye: I might call this `DEV_FUNCTIONS_SERVER` or similar instead, since (as I understand it) that's what value will be stored in this environment variable.

PR #145, opened 2019-10-29
Use PreconditionFailed and fix user agent strings bug

jacoblurye:
This PR fixes an unsafe `.split('/')` operation introduced in #142 that is currently causing all requests from browsers to the API to fail.

Also, this updates HTTP responses based on how out-of-date the requesting CLI is.

### DIFF
--- a/cidc_api/auth.py
+++ b/cidc_api/auth.py
@@ -7,7 +7,7 @@ import requests
 from eve.auth import TokenAuth
 from jose import jwt
 from flask import _request_ctx_stack, request, current_app as app
-from werkzeug.exceptions import Unauthorized, BadRequest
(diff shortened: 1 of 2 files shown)


cidc_api/auth.py
```diff
@@ -7,7 +7,12 @@
 from eve.auth import TokenAuth
 from jose import jwt
 from flask import _request_ctx_stack, request, current_app as app
-from werkzeug.exceptions import Unauthorized, BadRequest
+from werkzeug.exceptions import (
+    Unauthorized,
+    BadRequest,
+    PreconditionFailed,
+    InternalServerError,
```
curlup: unneeded import?

PR #148, opened 2019-10-29
More human friendly "Signature has expired" error reporting.

curlup:


cidc_api/auth.py
```diff
@@ -257,7 +257,7 @@ def decode_id_token(self, token: str, public_key: dict) -> dict:
                 options={"verify_at_hash": False},
             )
         except jwt.ExpiredSignatureError as e:
-            raise Unauthorized(str(e))
+            raise Unauthorized(f"{e}. Try log in again with cidc login.")
```
jacoblurye:
```suggestion
            raise Unauthorized(f"{e} Obtain a new login token from the CIDC Portal, then try logging in again.")
```

PR #156, opened 2019-11-06
Generate templates on startup

jacoblurye:
Now, the API generates empty schemas templates on application startup and serves them from the `/info/templates/...` endpoint.

Accompanying UI changes to come.

### DIFF
--- a/cidc_api/services/info.py
+++ b/cidc_api/services/info.py
@@ -1,6 +1,5 @@
 """Endpoints providing info related to this API"""
 import os
-import re
 
 from flask import Blueprint, jsonify, current_app as app, send_file
 from werkzeug.exceptions import NotFound, BadRequest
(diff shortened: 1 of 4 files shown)


cidc_api/services/info.py
```diff
@@ -30,3 +34,32 @@ def manifests():
...
+
+@info_api.route("templates/<template_family>/<template_type>", methods=["GET"])
+def templates(template_family, template_type):
+    """
+    Return the empty Excel template file for the given 
+    `template_family` (e.g., manifests, metadata) and 
+    `template_type` (e.g., pbmc, olink).
+    """
+    # Check that both strings are alphabetic
+    if not alphabetic.match(template_family):
```
curlup:
```
%%timeit a = re.compile(r"^[a-zA-Z]*$")
  a.match('rnaseq')
313 ns ± 8.54 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

%timeit 'rnaseq'.isalpha()
57.5 ns ± 0.459 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

jacoblurye: No kidding!

PR #159, opened 2019-11-07
Update response to set extra_metadata=None as default

roshni-b:
* Updated response to set extra_metadata = None as default
* Update WES upload test to assert that extra_metadata is None, while Olink upload test asserts that extra_metadata is a dict

### DIFF
--- a/tests/services/test_ingestion.py
+++ b/tests/services/test_ingestion.py
@@ -415,6 +415,7 @@ def test_upload_wes(app_no_auth, test_user, db_with_trial_and_user, db, monkeypa
     assert "url_mapping" in res.json
     url_mapping = res.json["url_mapping"]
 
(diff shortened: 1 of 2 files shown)


tests/services/test_ingestion.py
```diff
@@ -413,9 +413,13 @@ def test_upload_wes(app_no_auth, test_user, db_with_trial_and_user, db, monkeypa
     )
     assert res.json
     assert "url_mapping" in res.json
-
     url_mapping = res.json["url_mapping"]
 
+    # WES assay does have any extra_metadata files
```
jacoblurye:
```suggestion
    # WES assay does not have any extra_metadata files, but its (and every assay's) response 
    # should have an extra_metadata field.
```

PR #161, opened 2019-11-12
Send email alerting successful assay/manifest upload

jacoblurye:
This PR adds:
* the email template `new_upload_alert` to `emails.py`
* functionality for alerting upload success to the `AssayUploads` and `ManifestUploads` models
* a fix for a bug in `AssayUploadStatus`, where an assay upload in the `started` state could skip ahead to `merge-completed` before passing through `upload-completed`.

Next, trigger email alerts from the `ingest_upload` cloud function on success.

### DIFF
--- a/cidc_api/models.py
+++ b/cidc_api/models.py
@@ -639,7 +639,7 @@ class AssayUploads(CommonColumns, UploadForeignKeys):
         return upload
 
     @with_default_session
-    def ingestion_success(self, session):
+    def ingestion_success(self, session, commit=False):
         """Set own status to reflect successful merge and trigger email notifying CIDC admins."""
         # Do status update if the transition is valid
         if not AssayUploadStatus.is_valid_transition(
@@ -649,7 +649,9 @@ class AssayUploads(CommonColumns, UploadForeignKeys):
                 f"Cannot declare ingestion success given current status: {self.status}"
             )
         self.status = AssayUploadStatus.MERGE_COMPLETED.value
-        session.commit()
+
+        if commit:
+            session.commit()
 
         self.alert_upload_success()
 
--- a/cidc_api/models.py
+++ b/cidc_api/models.py
@@ -482,6 +482,7 @@ class ManifestUploads(CommonColumns, UploadForeignKeys):
         gcs_xlsx_uri: str,
         session: Session,
         commit: bool = True,
+        send_email: bool = False,
     ):
         """Create a new ManifestUpload for the given trial manifest patch."""
         assert (
@@ -500,6 +501,9 @@ class ManifestUploads(CommonColumns, UploadForeignKeys):
         session.add(upload)
         if commit:
             session.commit()
+
+        if send_email:
+            upload.alert_upload_success()
 
         return upload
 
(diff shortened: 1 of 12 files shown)


cidc_api/models.py
```diff
@@ -455,6 +455,14 @@ def trial(cls):
     # The object URI for the raw excel form associated with this upload
     gcs_xlsx_uri = Column(String, nullable=False)
 
+    def alert_upload_success(self):
+        """Send an email notification that an upload has succeeded."""
+        # (import these here to avoid a circular import error)
+        import emails, gcloud_client
```
curlup: why are you using an import here and not in the head of the module?

cidc_api/models.py
```diff
@@ -455,6 +455,14 @@ def trial(cls):
...
+
+        # Send admin notification email
+        gcloud_client.send_email(**emails.new_upload_alert(self))
```
curlup: also, is there a reason not to import and use `gcloud_client` in the `email` module directly?

cidc_api/models.py
```diff
@@ -624,6 +638,23 @@ def find_by_id_and_email(cls, id, email, session):
             return None
         return upload
 
+    @with_default_session
+    def ingestion_success(self, session, commit=False):
```
curlup:
```suggestion
    def ingestion_success(self, session: Session, commit: bool = False):
```

curlup: I would also suggest review/unification of default values for commit across models methods.  
Seems that it's somewhat a hodgepodge.

cidc_api/models.py
```diff
@@ -624,6 +638,23 @@ def find_by_id_and_email(cls, id, email, session):
...
+        """Set own status to reflect successful merge and trigger email notifying CIDC admins."""
+        # Do status update if the transition is valid
+        if not AssayUploadStatus.is_valid_transition(
```
curlup: with `self` being `AssayUploads` I think it might be more appropriate to call this as an instance method.

cidc_api/services/ingestion.py
```diff
@@ -287,6 +287,9 @@ def upload_manifest(
         session=session,
     )
 
+    # Manifest has been uploaded successfully
+    manifest_upload.alert_upload_success()
```
curlup: nit: as this `alert_upload_success` in case of assay upload is called within a model method, I would consider moving this call here to the `manifest.create` - as it `commit`s the session and thus should issue that email in a consistent way with assay uploads.

cidc_api/models.py
jacoblurye: `is_valid_transition` doesn't exist as an instance method on `AssayUploads`, only as a static method on the `AssayUploadStatus` enum.

cidc_api/models.py
jacoblurye: Nope! I can refactor.

cidc_api/models.py
jacoblurye: The `emails` module imports models from this file (though only for use in type hints), so doing that would lead to a circular import.

cidc_api/models.py
curlup:
```
In [4]: class What():
   ...:     ever = "nooo"
   ...:     @classmethod
   ...:     def yeah(cls):
   ...:         return cls.ever
   ...:

In [5]: What().yeah()
Out[5]: 'nooo'
```

cidc_api/models.py
```diff
@@ -482,6 +482,7 @@ def create(
         gcs_xlsx_uri: str,
         session: Session,
         commit: bool = True,
+        send_email: bool = False,
```
curlup: shouldn't default be `True`?

jacoblurye: Since the email alert is an unexpected side-effect of this method, my preference is towards `False`.

PR #163, opened 2019-11-14
Olink metadata migration

jacoblurye:
This PR bumps the CIDC schemas package version and adds a metadata migration from v0.10.0 of the data model to v0.10.2.


migrations/versions/ff3141aecdd4_.py
```diff
@@ -0,0 +1,80 @@
...
+branch_labels = None
+depends_on = None
+
+is_testing = os.environ.get("TESTING")
+
+session = Session(bind=op.get_bind())
+storage_client = storage.Client() if not is_testing else None
+
+
+def _do_metadata_migration(metadata_migration: Callable[[dict], MigrationResult]):
```
jacoblurye: I wonder if this code is generic enough that it should live in a separate module somewhere. It's possible we'll be able to use it directly for future metadata migrations.

migrations/versions/ff3141aecdd4_.py
```diff
@@ -0,0 +1,80 @@
...
+        trial.metadata_json = migration.result
+
+        # Update the relevant downloadable files and GCS objects
+        for old_gcs_uri, artifact in migration.file_updates.items():
+            # Update the downloadable file associated with this blob
+            df = (
+                session.query(DownloadableFiles)
+                .filter_by(object_url=old_gcs_uri)
+                .with_for_update()
+                .one()
```
curlup: shouldn't this be `.all()` too?

migrations/versions/ff3141aecdd4_.py
```diff
@@ -0,0 +1,80 @@
...
+            )
+            for column, value in artifact.items():
+                if hasattr(df, column):
+                    setattr(df, column, value)
```
curlup: what about `additional_metadata`?

migrations/versions/ff3141aecdd4_.py
```diff
@@ -0,0 +1,80 @@
...
+
+            # If the GCS URI has changed, rename the blob
+            new_gcs_uri = artifact["object_url"]
+            if old_gcs_uri != new_gcs_uri:
+                bucket = storage_client.get_bucket(GOOGLE_DATA_BUCKET)
```
curlup: should we maybe also move them in the upload bucket as well?

migrations/versions/ff3141aecdd4_.py
curlup: Yeah, would be nice.

setup.py
```diff
@@ -14,6 +14,6 @@
     packages=["cidc_api.config"],
     py_modules=["cidc_api.models", "cidc_api.gcloud_client", "cidc_api.emails"],
     url="https://github.com/CIMAC-CIDC/cidc_api-gae",
-    version="0.9.4",
+    version="0.9.5",
```
curlup: Tip: I'm generally trying to line up api major.minor versions with schemas. Makes it easier to track.

migrations/versions/ff3141aecdd4_.py
jacoblurye: As `downloadable_files.object_url` is constrained to be unique, we are guaranteed to get only one record.

migrations/versions/ff3141aecdd4_.py
jacoblurye: good point! added.

cidc_api/utils/migrations.py
```diff
@@ -0,0 +1,159 @@
...
+
+def run_metadata_migration(metadata_migration: Callable[[dict], MigrationResult]):
+    """Migrate trial metadata, upload job patches, and downloadable files according to `metadata_migration`"""
+    session = Session(bind=op.get_bind())
+
+    # Initialize an empty list of pieces of work to do on GCS resources
+    gcs_tasks = RollbackableQueue()
+
+    # Migrate all trial records
+    trials: List[TrialMetadata] = session.query(TrialMetadata).with_for_update().all()
```
curlup: this might potentially block the whole table for the time of this operation - not sure how `with_for_update` and `all` work together.

Anyways it might be okay to do that, if estimates on how long things might take are not too big.
But I think it worth checking.

cidc_api/utils/migrations.py
```diff
@@ -0,0 +1,159 @@
...
+        migration = metadata_migration(trial.metadata_json)
+
+        # Update the trial metadata object
+        trial.metadata_json = migration.result
+
+        # Update the relevant downloadable files and GCS objects
+        for old_gcs_uri, artifact in migration.file_updates.items():
+            # Update the downloadable file associated with this blob
+            df: DownloadableFiles = (
+                session.query(DownloadableFiles)
```
curlup: Isn't there a way to do update right away, without `with_for_update` locking here?
Seems that unlike as with `TrialMetadata` here we don't rely on current prop values from DF record, to set new ones.

cidc_api/utils/migrations.py
```diff
@@ -0,0 +1,159 @@
...
+                    lambda: rename_gcs_blob(
+                        GOOGLE_DATA_BUCKET, new_gcs_uri, old_gcs_uri
+                    ),
+                )
+                gcs_tasks.schedule(renamer)
+
+    # Migrate all assay upload successes
+    successful_assay_uploads: List[AssayUploads] = session.query(
+        AssayUploads
+    ).filter_by(status=AssayUploadStatus.MERGE_COMPLETED.value).with_for_update().all()
```
curlup: btw, same `with_for_update().all()` here doesn't look that concerning as the first one - here we're updating only  `MERGE_COMPLETED` assay uploads, that are probably not in the use by any other process, afaik.

cidc_api/utils/migrations.py
```diff
@@ -0,0 +1,159 @@
...
+
+        # Update the GCS URIs of files that were part of this upload
+        old_file_map = upload.gcs_file_map
+        new_file_map = {}
+        for (
+            old_upload_uri,
+            old_target_uri,
+            artifact_uuid,
+        ) in upload.upload_uris_with_data_uris_with_uuids():
+            upload_timestamp = old_upload_uri[len(old_target_uri) + 1 :]
```
curlup: nit: I would abstract that into a helper function. at some point I wanted to add extensions to those upload filenames as say `uploads/staging_0/olink/chip_1chip/assay_raw_ct.xlsx/2019-10-07T20:00:40.903770.xlsx` so they will be opened right away

cidc_api/utils/migrations.py
```diff
@@ -0,0 +1,159 @@
...
+            if old_target_uri in migration.file_updates:
+                new_target_uri = migration.file_updates[old_target_uri]["object_url"]
+                if old_gcs_uri != new_target_uri:
```
curlup: shouldn't that be `old_target_uri` instead of `old_gcs_uri`

cidc_api/utils/migrations.py
```diff
@@ -0,0 +1,159 @@
...
+                    new_upload_uri = "/".join([new_target_uri, upload_timestamp])
+                    renamer = PieceOfWork(
```
curlup: wait, I though we decided not to move stuff in the upload bucked?

cidc_api/utils/migrations.py
```diff
@@ -0,0 +1,159 @@
...
+        migration = metadata_migration(upload.metadata_patch)
+
+        # Update the metadata patch
+        upload.metadata_patch = migration.result
+
+    # Attempt to make GCS updates
+    gcs_tasks.run_all()
+
+    # Update the database if everything else has succeeded
+    session.commit()
```
curlup: so I feel like all those `with_for_update().all` might hold locks up until here, which seems like a lot of time / work in between.

curlup: it seems there should be a rollback section somewhere, in case something goes wrong with gcs copying or any other step. right?

migrations/versions/ff3141aecdd4_.py
```diff
@@ -0,0 +1,37 @@
...
+branch_labels = None
+depends_on = None
+
+is_testing = os.environ.get("TESTING")
+
+
+def upgrade():
+    """Update Olink's assay_raw_ct artifact data format to CSV"""
+    # Don't run this migration on the test database
+    if is_testing:
```
curlup: not sure why it shouldn't. because it's supposed to be empty?

jacoblurye: It _will_ be empty, but the reason why is that this migration accesses GCS resources that Travis does not have access to, so running it in CI would fail.

That being said, this detail should probably be encapsulated inside `run_metadata_migration`.

cidc_api/utils/migrations.py
jacoblurye: I know we went back and forth on that. Here's my reasoning: migrating the `assay_uploads`  URIs and metadata patches without updating GCS would mean effectively losing the original files associated with a given upload from the system's perspective; since we decided to migrate `assay_uploads` metadata and URIs, I decided to also move stuff in the upload bucket.

curlup:
Couple of tests suggestions
* sql rollback on gcs mv issue
* sql rollback on sql issue
* gcs rollback on sql issue

cidc_api/utils/migrations.py
jacoblurye: I'd rather do this in a separate PR, since this helper function (and an associated helper for building URLs) should be used everywhere functionality like this exists in the API code.

cidc_api/utils/migrations.py
jacoblurye: (we discussed offline and concluded that since only cloud functions would potentially get blocked by this lock, and cfns are background tasks, we're ok with this for now)

cidc_api/utils/migrations.py
jacoblurye: https://github.com/CIMAC-CIDC/cidc-api-gae/pull/163#discussion_r348141441

cidc_api/utils/migrations.py
```diff
@@ -47,10 +48,28 @@ def run_all(self):
...
+@contextmanager
+def migration_session():
+    session = Session(bind=op.get_bind())
+
+    try:
+        yield session
+    except:
+        session.rollback()
+        raise
+    finally:
```
curlup: > The finally clause runs whether or not the try statement produces an exception.

I think an `else` should be used here.

jacoblurye: Actually, according to the SQLAlchemy docs, we want:
```python
    try:
        yield session
        session.commit()
    except:
        session.rollback()
        raise
    finally:
        session.close()
```
Whoops 😅

cidc_api/utils/migrations.py
```diff
@@ -42,39 +43,67 @@ def run_all(self):
...
+    task_queue = RollbackableQueue()
 
     try:
-        yield session
+        yield session, task_queue
+        session.commit()
     except:
         session.rollback()
+        if task_queue:
+            task_queue.rollback()
```
curlup: in case this `.rollback` call fails/raises, will we get both exceptions in the logs ?

tests/utils/test_migrations.py
```diff
@@ -24,3 +40,74 @@ def test_rollbackable_queue():
...
+        mock_session.commit.reset_mock()
+        mock_session.rollback.reset_mock()
+        mock_session.close.reset_mock()
+
+    # GCS failure
+    rename_gcs_obj.side_effect = Exception("gcs failure")
+
+    with pytest.raises(Exception, match="gcs failure"):
+        run_metadata_migration(mock_migration)
+    rename_gcs_obj.assert_called_once()
```
curlup: shouldn't it try to rename stuff back here?

cidc_api/utils/migrations.py
```diff
@@ -65,14 +65,21 @@ def migration_session():
...
+        print("Running SQL rollback.")
         session.rollback()
         if task_queue:
             try:
+                print("Running GCS rollback.")
                 task_queue.rollback()
+                print("GCS rollback succeeded.")
             except Exception as e:
                 print(f"GCS rollback failed: {e.__class__}: {e}")
+        print("SQL rollback failed.")
```
curlup: wait, is rollback failed here?

jacoblurye: Nope! Good catch.

cidc_api/utils/migrations.py
```diff
@@ -0,0 +1,229 @@
...
+                        f"Encountered GCS upload bucket artifact URI to update: {old_upload_uri}"
+                    )
+                    new_upload_uri = "/".join([new_target_uri, upload_timestamp])
+                    renamer = PieceOfWork(
+                        lambda: rename_gcs_blob(
+                            GOOGLE_UPLOAD_BUCKET, old_upload_uri, new_upload_uri
+                        ),
+                        lambda: rename_gcs_blob(
+                            GOOGLE_UPLOAD_BUCKET, new_upload_uri, old_upload_uri
+                        ),
```
curlup:
```
In [1]: l=[]

In [2]: for i in range(5):
   ...:     l.append(lambda: print(i))
   ...:

In [3]: for la in l:
   ...:     la()
   ...:
4
4
4
4
4
```

jacoblurye: 😞

PR #165, opened 2019-11-20
template names allow for underscore

curlup:


tests/services/test_info.py
```diff
@@ -48,8 +48,15 @@ def test_templates(app_no_auth):
     assert res.json["_error"]["message"] == "Invalid template family: .."
 
     res = client.get(f"{INFO_ENDPOINT}/templates/manifests/pbmc123")
+    assert res.status_code == 404
+    assert (
+        res.json["_error"]["message"]
+        == "No template found for the given template family and template type"
+    )
+
```
jacoblurye: Can you also add a test demonstrating that a template name with underscore works now?

PR #184, opened 2019-12-12
Raise Eve pagination default

jacoblurye:
See the entry in https://docs.python-eve.org/en/stable/config.html on `PAGINATION_LIMIT` and `PAGINATION_DEFAULT` - raising our `PAGINATION_DEFAULT` will make our lives a bit easier, at least until we get proper use of paginated endpoints set up in the portal UI.


cidc_api/config/settings.py
```diff
@@ -68,6 +68,5 @@
 ITEM_METHODS = ["GET", "PATCH"]
 CACHE_CONTROL = "no-cache"
 DOMAIN = get_DOMAIN()
-PAGINATION_DEFAULT = 100
-PAGINATION_LIMIT = 200
```
curlup: wouldn't removing the limit setting make the limit to be equal to the limit default which is 50 acc to docs?

jacoblurye: Missed the detail that `PAGINATION_LIMIT` defaults to 50.

PR #198, opened 2020-01-10
Bump schemas version and publish manifest ID, not trial ID, on manifest upload

jacoblurye:
Publishing the manifest upload ID to the `patient_sample_update` topic is required to use `unprism.derive_files` in the `generate_csvs` cloud function.

### DIFF
--- a/cidc_api/services/ingestion.py
+++ b/cidc_api/services/ingestion.py
@@ -308,8 +308,8 @@ def upload_manifest(
         send_email=True,
     )
 
-    # Publish that this trial's metadata has been updated
-    gcloud_client.publish_patient_sample_update(trial.trial_id)
+    # Publish that a manifest upload has been received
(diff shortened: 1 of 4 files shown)


cidc_api/services/ingestion.py
```diff
@@ -308,8 +308,8 @@ def upload_manifest(
         send_email=True,
     )
 
-    # Publish that this trial's metadata has been updated
-    gcloud_client.publish_patient_sample_update(trial.trial_id)
+    # Publish that a manifest upload has been received
+    gcloud_client.publish_patient_sample_update(manifest_upload.trial_id)
```
curlup: I'm lost 
```
def publish_patient_sample_update(manifest_upload_id: int):
```
vs this here 
```
publish_patient_sample_update(manifest_upload.trial_id)
```
So should that be a `trial_id` or `manifest_upload_id` ?

jacoblurye: typo!

PR #207, opened 2020-01-27
Consolidate assay_uploads and manifest_uploads into unified upload_jobs table

jacoblurye:
This PR aims to pay off some technical debt: currently, manifest uploads and assay/analysis uploads are tracked in different tables in the database. Some minor modifications to the `assay_uploads` table makes it possible to track manifest uploads there, too.

This update would require changes throughout the system, which it makes it possibly higher risk than it's worth:
* [x] the CLI needs to make a PATCH request against `/upload_jobs`, not `/assay_uploads`, at the end of assay/analysis ingestion: https://github.com/CIMAC-CIDC/cidc-cli/pull/109
* [x] cloud functions needs to query the database using the `UploadJobs` model instead of `AssayUploads` and `ManifestUploads`: https://github.com/CIMAC-CIDC/cidc-cloud-functions/pull/108


cidc_api/utils/migrations.py
```diff
@@ -88,31 +90,66 @@ def migration_session():
...
 def _select_trials(session: Session) -> List[TrialMetadata]:
     return session.query(TrialMetadata).with_for_update().all()
 
 
-def _select_successful_assay_uploads(session: Session) -> List[AssayUploads]:
+class AssayUploads(CommonColumns):
+    """This model no longer exists in cidc_api.models, but a partial model is required for migrations."""
+
+    __tablename__ = "assay_uploads"
+    status = Column(String)
```
curlup: I think this breaks line :218 down below

PR #215, opened 2020-02-04
Foreign key refactor

jacoblurye:
@curlup and I noticed that for every 1 faceted search filter request to the API, the Eve/SQLAlchemy was making an additional `SELECT` query against the DB for *every single downloadable file record* returned by the initial faceted search query. Turns out Eve/SQLAlchemy was doing an application-level join on the `downloadable_files.trial_id` foreign key.

These changes simplify the way we specify foreign key constraints and should prevent Eve/SQLAlchemy from making those additional queries.

related UI changes: https://github.com/CIMAC-CIDC/cidc-ui/pull/199


cidc_api/models.py
```diff
@@ -74,30 +73,35 @@ class ExtraDataTypes(EnumBaseClass):
 ]
 
 
+class FKResourceConfig(ResourceConfig):
+    def _get_column_fields(self):
+        """
+        Eve-SQLAlchemy's built-in ResourceConfig generator doesn't include foreign keys
+        without relationships in the schema config, but we want to allow this, so
+        we override this method: https://github.com/pyeve/eve-sqlalchemy/blob/22bdd96fe5b461fbdfe4a4dfdc394569f7e9b8b4/eve_sqlalchemy/config/resourceconfig.py#L157
```
curlup: we override this method
```suggestion
        we override this method: https://github.com/pyeve/eve-sqlalchemy/blob/22bdd96fe5b461fbdfe4a4dfdc394569f7e9b8b4/eve_sqlalchemy/config/resourceconfig.py#L157
        by removing `len(f.expression.foreign_keys) == 0)` filter from it
```

cidc_api/models.py
```diff
@@ -663,8 +677,8 @@ class DownloadableFiles(CommonColumns):
     upload_type = Column(String, nullable=False)
     md5_hash = Column(String, nullable=True)
     crc32c_hash = Column(String, nullable=True)
-    trial_id = Column(String, ForeignKey("trial_metadata.trial_id"), nullable=False)
-    trial = relationship(TrialMetadata, foreign_keys=[trial_id])
+    trial_id = Column(String, nullable=False)
+    # trial = relationship(TrialMetadata, foreign_keys=[trial_id])
```
curlup:
```suggestion
```

jacoblurye:
I changed all foreign key references, not just the ones in `DownloadableFiles`, because without those changes, we can expect the same application-side join behavior for those, too - for example, when the portal makes a `GET /permissions` request. I also think the models are a bit easier to understand this way.

Any arguments against this, @curlup?

PR #217, opened 2020-02-10
Prod release

curlup:
Is it ready? 

What's the order wrt CFns?


jacoblurye:
@curlup these changes appear to already be on `production` (see, for example: https://github.com/CIMAC-CIDC/cidc-api-gae/blob/production/cidc_api/models.py). Seems like some squashing / commit-hash-renaming weirdness has led to the diff visible in this PR. Feel free to merge if that might clear the discrepancy up.

PR #220, opened 2020-02-12
Use absolute, not relative, imports in cidc_api.config

jacoblurye:
Randomly, the staging app engine instance has started throwing the following error on startup:
```
File "/srv/cidc_api/config/settings.py", line 7, in <module> from . import db ImportError: attempted relative import with no known parent package
```
Here's a hot fix that should clear up this error up while we investigate why it started happening.

### DIFF
--- a/cidc_api/config/db.py
+++ b/cidc_api/config/db.py
@@ -1,6 +1,6 @@
 from os import environ
 
-from .secrets import get_secrets_manager
+from cidc_api.config.secrets import get_secrets_manager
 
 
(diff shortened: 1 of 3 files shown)


cidc_api/config/db.py
```diff
@@ -1,6 +1,6 @@
 from os import environ
 
-from .secrets import get_secrets_manager
```
curlup: Did these also errored or is it just in case?

jacoblurye: Just in case.

PR #228, opened 2020-02-24
Add tests for resource / item endpoints in preparation for Eve-SQLAlchemy phase-out

jacoblurye:
This PR does a couple things:
* Adds smoke tests for resource / item endpoints currently generated by Eve-SQLAlchemy. Includes checks for different HTTP methods and for pagination / sorting / filtering.
* Fixes some bugs discovered by those tests.
* Removes `eve-swagger` dependency. We weren't using these generated docs.
* Removes an unused function for generating signed GCS download URLs.


tests/test_api.py
```diff
@@ -0,0 +1,386 @@
...
+#   `model` (required): the SQLAlchemy model for this resource.
+#   `allowed_methods` (required): the HTTP methods this resource supports.
+#   `POST_setup`: a list of other resources to add to the database before POSTing this resource.
+#   `PATCH_json` (required if "PATCH" in `allowed_methods`): a JSON patch update for this resource.
+#   `filters`: a dictionary containing two entries representing possible filter queries:
+#       `empty`: a query filter that should return empty results.
+#       `one`: a query filter that should return exactly one result.
+#   `additional_records`: a list of JSON instances of this resource to insert before testing pagination.
+#   `mocks`: a list of functions that accept pytest's `monkeypatch` as their argument.
+resource_requests = {
```
curlup: I would think on somehow splitting that huge object that represents not only multiple tests, but multiple test for multiple models, which is too much I think.

That would also allow to include `json` instance of a resource directly into every such record, brining them together as something like
```
resource_requests = {}
resource_requests['new_users']={
        "json": {
            **new_user,
            "id": TEST_RECORD_ID,
            "role": CIDCRole.ADMIN.value,
            "approval_date": datetime.now(),
        },
        "model": Users,
        "allowed_methods": {"POST", "PATCH", "GET"},
        "PATCH_json": {"role": CIDCRole.CIMAC_USER.value},
...
```

tests/test_api.py
```diff
@@ -0,0 +1,386 @@
...
+        assert base[key] == value or isinstance(value, datetime)
+
+
+def setup_mocks(config, monkeypatch):
+    if "mocks" in config:
+        for mock in config["mocks"]:
+            mock(monkeypatch)
+
+
+@pytest.mark.parametrize("resource, config", resource_requests.items())
```
curlup: i'd add `ids` fn that will give human readable descriptions to the parameter set:
```suggestion
@pytest.mark.parametrize("resource, config", resource_requests.items(), ids=lambda x: x if isinstance(x, str) else x['model'])
```

tests/test_api.py
```diff
@@ -0,0 +1,386 @@
...
+        "additional_records": [
+            {**user, "id": 2, "email": "foo@bar.com"},
+            {**user, "id": 3, "email": "fizz@buzz.com"},
+        ],
+    },
+    "trial_metadata": {
+        "json": trial_metadata,
+        "model": TrialMetadata,
+        "allowed_methods": {"POST", "PATCH", "GET"},
+        "lookup_field": "trial_id",
```
curlup: I would add a documentation for that `lookup_field` in the header comment block

tests/test_api.py
```diff
@@ -0,0 +1,386 @@
...
+
+@pytest.mark.parametrize("resource, config", resource_requests.items())
+def test_item_patch(
+    resource, config, app_with_admin_user, db_with_records, monkeypatch
+):
+    client = app_with_admin_user.test_client()
+    setup_mocks(config, monkeypatch)
+
+    # Try to update the resource
+    lookup = config["json"].get(config.get("lookup_field") or "id")
```
curlup: seems repetitive - extract a helper?

tests/test_api.py
```diff
@@ -0,0 +1,386 @@
...
+    assert len(one_response.json["_items"]) == 1
+    item = one_response.json["_items"][0]
+    assert_dict_contains(item, config["json"])
+
+    empty_response = client.get(resource, query_string=config["filters"]["empty"])
+    assert empty_response.status_code == 200
+    assert empty_response.json["_items"] == []
+
+
+@pytest.mark.parametrize("resource, config", resource_requests.items())
```
curlup: maybe move this filtering 
```
if "additional_records" not in config:
        return
```
into parametrize call so no dummy tests will be run
```suggestion
@pytest.mark.parametrize("resource, config", [r,c for r,c in resource_requests.items() if if "additional_records" in c])
```

tests/test_api.py
```diff
@@ -0,0 +1,386 @@
...
+    db_with_records.commit()
+
+    setup_mocks(config, monkeypatch)
+    client = app_with_admin_user.test_client()
+
+    # Check that max_results = 1 returns only one result
+    response = client.get(resource, query_string={"max_results": 1})
+    assert response.status_code == 200
+    assert len(response.json["_items"]) == 1
+    assert response.json["_items"][0]["id"] == 1
```
curlup:
```suggestion
    assert response.json["_items"][0]["id"] == TEST_RECORD_ID
```

tests/test_api.py
jacoblurye: Oh, cool feature! Will do.

tests/test_api.py
jacoblurye: The auto-generated parameter labels seem friendly enough, IMO:
```
tests/test_api.py::test_resource_post[new_users-config0]
tests/test_api.py::test_resource_post[users-config1]
tests/test_api.py::test_resource_post[trial_metadata-config2]
tests/test_api.py::test_resource_post[downloadable_files-config3]
tests/test_api.py::test_resource_post[permissions-config4]
...
```

tests/test_api.py
```diff
@@ -0,0 +1,386 @@
...
+#       `one`: a query filter that should return exactly one result.
+#   `additional_records`: a list of JSON instances of this resource to insert before testing pagination.
+#   `mocks`: a list of functions that accept pytest's `monkeypatch` as their argument.
+resource_requests = {
+    "new_users": {"json": new_user, "model": Users, "allowed_methods": {"POST"}},
+    "users": {
+        "json": user,
+        "model": Users,
+        "allowed_methods": {"POST", "PATCH", "GET"},
+        "PATCH_json": {"role": CIDCRole.CIMAC_USER.value},
```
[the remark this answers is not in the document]
jacoblurye: Currently, the tests just check that the database record contains the `PATCH_json` after the PATCH request is made. Not exhaustive, but maybe sufficient?

tests/test_api.py
```diff
@@ -0,0 +1,386 @@
...
+
+TEST_RECORD_ID = 1
+new_user = {"email": "test-admin@email.com"}
+user = {
+    **new_user,
+    "id": TEST_RECORD_ID,
+    "role": CIDCRole.ADMIN.value,
+    "approval_date": datetime.now(),
+}
+trial_metadata = {"id": TEST_RECORD_ID, "trial_id": "foo", "metadata_json": {}}
```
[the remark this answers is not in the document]
jacoblurye: While I agree that this *should* be the behavior of our API, it isn't. There is no pre-insert hook that runs to validate the `metadata_json` field when a new `trial_metadata` record is POSTed to the API (currently only admins are allowed to POST against this resource).

tests/test_api.py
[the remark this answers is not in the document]
curlup: sure

tests/test_api.py
curlup: right, that's friendly !

PR #232, opened 2020-02-26
Add option to publish to artifact_upload topic on downloadable_file creation

jacoblurye:
This is intended for use in cloud functions.

### DIFF
--- a/cidc_api/models.py
+++ b/cidc_api/models.py
@@ -689,6 +689,11 @@ class DownloadableFiles(CommonColumns):
     clustergrammer = Column(JSONB, nullable=True)
     ihc_combined_plot = Column(JSONB, nullable=True)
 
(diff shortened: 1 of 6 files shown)


cidc_api/models.py
```diff
@@ -689,6 +689,11 @@ class DownloadableFiles(CommonColumns):
     clustergrammer = Column(JSONB, nullable=True)
     ihc_combined_plot = Column(JSONB, nullable=True)
 
+    def publish_artifact_upload(self):
+        from gcloud_client import publish_artifact_upload
```
curlup: Would be cool if you could resolve that within this PR

jacoblurye: Agreed. (`this` being the underlying circular import issue this skirts around.)

PR #233, opened 2020-02-27
Fix get_secret_manager import in evesettings.py

jacoblurye:


cidc_api/config/evesettings.py
```diff
@@ -31,7 +31,7 @@
 # fails with a credentials-related error unless this is explicitly
 # set.
 if not os.environ.get("GOOGLE_APPLICATION_CREDENTIALS") and not TESTING:
-    secret_manager = secrets.get_secret_manager()
+    secret_manager = get_secret_manager(TESTING)
```
curlup: Not sure that I follow - `not TESTING` 2 lines above and passing `TESTING` here

jacoblurye: `get_secret_manager` used to take a required positional argument `is_testing: bool`, but I'll remove that requirement.

PR #245, opened 2020-04-23
Remove Eve-SQLAlchemy

jacoblurye:
This PR entirely removes our Eve-SQLAlchemy dependency. Here's an overview of the new project structure and changes to individual files, with bold text representing an Eve feature that's been replaced. If a file isn't mentioned, that's because it wasn't changed.

- _`cidc_api.app`_ - instantiates and configures the `Flask` application object and adds **HTTP error handlers**.
- _`cidc_api.config`_ - modules related to application setup and configuration
  - `settings` - recombination of `evesettings` and `settings`
  - `db` - our SQLAlchemy `declarative_base` instance is now initialized here instead of `cidc_api.models`
- _`cidc_api.models`_ - SQLAlchemy model definitions and converters
  - `models` - SQLAlchemy models defining the CIDC database schema, unchanged aside from the introduction of **common controller methods like `insert`, `update`, `list`, `delete`, and `count`** to assist in implementing REST logic. The **`list` method implements basic pagination, filtering, and sorting support**.
  - `schemas` - [marshmallow-sqlalchemy](https://marshmallow-sqlalchemy.readthedocs.io/) schemas for **validating and converting to and from JSON representations of the SQLAlchemy models**.
- _`cidc_api.shared`_ - modules that are used across CIDC API endpoints, formerly stored at the top level in the `cidc_api` directory
  - `gcloud_client` - largely unchanged, but test coverage improved
  - `auth` - **removed dependency on Eve's `BearerAuth` class**, but authentication logic is largely the same. Now all endpoint authentication happens via the `requires_auth` decorator.
  - `rest_utils` - **Helpers (primarily decorators) for doing request / response JSON-SQLAlchemy conversions, item lookups, ETag-based concurrency controls, and pagination argument parsing.** Understanding these helpers is key to understanding the REST endpoint implementations.
- _`cidc_api.resources`_ - **REST endpoints for operating on CIDC resources**, plus the non-Eve endpoints that used to live in `cidc_api.services`. This is the meat and potatoes of the API.

The diff for this PR is basically impossible to parse, so I recommend pulling the code down and perusing it locally as a way to start reviewing it. Reading through the tests should help clarify intended functionality for any new code.

**Here's a PR with the required UI changes**: https://github.com/CIMAC-CIDC/cidc-ui/pull/219/files


jim-bo:
@jacoblurye I'm just going to add these as stream of consciousness...

Question 1: Is there anyway to force all endpoints have the auth decorator by default, while allowing for explicit exceptions if necessary? I think would reduce the chance of someone else making a security hole.

jacoblurye:
@jim-bo that should be possible, yes!

cidc_api/models/schemas.py
```diff
@@ -0,0 +1,77 @@
...
+    class Meta:
+        sqla_session = db.session
+        include_fk = True
+        load_instance = True
+
+
+class _ListMetadata(BaseSchema):
+    total = fields.Int(required=True)
+    # TODO: do we need these fields?
+    # page_num = fields.Int(required=True)
```
curlup, line 28: wouldn't those be useful for showing pagination in the UI?

cidc_api/models/schemas.py
```diff
@@ -0,0 +1,77 @@
...
+
+PermissionListSchema = _make_list_schema(PermissionSchema())
+
+
+class TrialMetadataSchema(BaseSchema):
+    class Meta(BaseSchema.Meta):
+        model = TrialMetadata
+
+
+TrialMetadataListSchema = _make_list_schema(TrialMetadataSchema())
```
curlup, line 77: I would expect for it to be possible to be generated in runtime with class-"magic" of python. In sorta the same way I believe Models work in sqlAlchemy. This is totally fine as it is, but if you'd like to go into more advanced python class inheritance whatever - that could be a good opportunity.

cidc_api/resources/downloadable_files.py
```diff
@@ -0,0 +1,137 @@
...
+
+    # Ensure user has permission to access this file
+    perms = Permissions.find_for_user(user.id)
+    if user.role != CIDCRole.ADMIN.value:
+        # Check for a permission matching this file's trial and assay
+        if not any(
+            perm.upload_type == file_record.upload_type
+            and perm.trial_id == file_record.trial_id
+            for perm in perms
+        ):
```
curlup, line 105: maybe it would be safer to have `Permissions` method that will only return perms for specific upload type and trial id? Than this (little bit) tricky thing here would become easier to read/understand and not to screw up (not that I'm saying it's screwed). Also getting from the DB only needed to check perms would probably work faster.

cidc_api/resources/downloadable_files.py
```diff
@@ -0,0 +1,137 @@
...
+    user = get_current_user()
+
+    if user.role == CIDCRole.ADMIN.value:
+        # Admins can facet on every trial or upload type
+        trial_ids = DownloadableFiles.get_distinct("trial_id")
+        upload_types = DownloadableFiles.get_distinct("upload_type")
+    else:
+        # Non-admins can only facet on what they have permission to view
+        perms = Permissions.find_for_user(user.id)
+        trial_ids = list({perm.trial_id for perm in perms})
```
curlup, line 134: why not `Permissions.for_user.distinct(trial_id | upload_type)` here as you did higher for admins?

cidc_api/resources/downloadable_files.py
```diff
@@ -0,0 +1,137 @@
...
+        upload_types = list({perm.upload_type for perm in perms})
+
+    return jsonify({"trial_id": trial_ids, "upload_type": upload_types})
```
curlup, line 137: This might be out of scope of this PR, but I think it might be better from UX point to return distinct pairs - trial_id+upload_type, so UI can then filter available upload_types, whenever user selects a trial and vice versa.

cidc_api/app.py
```diff
@@ -1,40 +1,51 @@
...
+from .config.db import init_db
+from .config.settings import SETTINGS
+from .shared.auth import validate_api_auth
+from .resources import register_resources
+
+app = Flask(__name__, static_folder=None)
+app.config.update(SETTINGS)
+
+# Enable CORS
+CORS(app, resources={r"*": {"origins": "*"}})
```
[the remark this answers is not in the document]
jacoblurye: Definitely - and this is always the configuration we've had 😓 How about werestrict this by by environment - i.e., `dev` allows localhost portal, `staging` allows staging portal, and `prod` allows prod portal?

cidc_api/models/models.py
```diff
@@ -173,25 +52,112 @@ def with_default_session(f):
...
 
+    @with_default_session
+    def insert(self, session: Session, commit: bool = True):
+        """Add the current instance to the session."""
+        session.add(self)
+        if commit:
+            session.commit()
+
+    @with_default_session
+    def update(self, session: Session, changes=None, commit: bool = True):
```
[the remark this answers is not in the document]
jacoblurye: Right now, the `lookup` decorator handles concurrency checks. Does that seem like the wrong place to you? https://github.com/CIMAC-CIDC/cidc-api-gae/blob/1b705ea6615c0385fc40d786156e96fab6b5ec73/cidc_api/shared/rest_utils.py#L81

cidc_api/models/models.py
```diff
@@ -534,10 +519,23 @@ class UploadJobs(CommonColumns):
         "upload_jobs_gcs_gcs_file_map_idx", gcs_file_map, postgresql_using="gin"
     )
 
+    @with_default_session
+    def update(self, session, changes=None, commit=True):
+        # If changes includes a status update, make sure that transition is valid
+        if (
+            changes
+            and changes.status
+            and not UploadJobStatus.is_valid_transition(self.status, changes.status)
```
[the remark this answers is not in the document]
jacoblurye: Better yet, I'll add an attribute validator https://docs.sqlalchemy.org/en/13/orm/mapped_attributes.html#simple-validators

cidc_api/models/schemas.py
jacoblurye: We don't currently use this in the UI implementation, so I'm not convinced it's useful.

cidc_api/resources/downloadable_files.py
jacoblurye: Want to make a ticket linking back to this suggestion? I'll hold off for now, but I like this idea.

cidc_api/models/schemas.py
jacoblurye: Not sure what exactly you're suggesting here, but the most important question to me is whether such a change would increase (or decrease) maintainability.

cidc_api/shared/auth.py
```diff
@@ -0,0 +1,322 @@
...
+def _extract_token() -> str:
+    """Extract an identity token from the current request's authorization header."""
+    auth_header = request.headers.get("Authorization")
+
+    if not auth_header or not auth_header.lower().startswith("bearer"):
+        raise Unauthorized(
+            "Authorization header must be set with structure 'Authorization: Bearer <id token>'"
+        )
+
+    id_token = auth_header.split(" ")[1]
```
curlup, line 133: totally nitpick: good practice would be, I think, return "400 bad request" in case of malformed auth header ((or `unauth`, but not `500 server error`), which this will do for something like `Authorization: Bearer` without following token

cidc_api/shared/auth.py
```diff
@@ -0,0 +1,322 @@
...
+
+    Returns:
+        dict: the decoded token as a dictionary.
+    """
+    try:
+        payload = jwt.decode(
+            token,
+            public_key,
+            algorithms=ALGORITHMS,
+            # TODO: is this what we want?
```
curlup, line 195: not sure about what this question means

cidc_api/shared/auth.py
```diff
@@ -0,0 +1,322 @@
...
+        str: the public key.
+    """
+    try:
+        header = jwt.get_unverified_header(token)
+    except jwt.JWTError as e:
+        raise Unauthorized(str(e))
+
+    # Get public keys from our Auth0 domain
+    jwks_url = f"https://{AUTH0_DOMAIN}/.well-known/jwks.json"
+    jwks = requests.get(jwks_url).json()
```
curlup, line 158: so we'll be doing one auth0 request for each user request, right? 
were we doing it like this before this PR? just asking

requirements.modules.txt
```diff
@@ -1,8 +1,6 @@
 flask~=1.1.1
 flask-sqlalchemy~=2.4.0
-werkzeug~=0.15.4 # required to fix an Eve-SQLAlchemy dependency issue
-pymongo==3.5.0 # required to fix an Eve-SQLAlchemy dependency issue
-eve-sqlalchemy~=0.7.0
```
curlup, line 5: 👍

tests/resources/test_permissions.py
```diff
@@ -0,0 +1,185 @@
...
+        other_user.insert()
+
+        # Create trial
+        TrialMetadata.create(TRIAL_ID, {})
+
+        # Create permissions
+        def create_permission(uid, assay):
+            Permissions(
+                granted_by_user=uid,
+                granted_to_user=uid,
```
curlup, line 47: nit: not with regard to this code, but shouldn't granting perms to a user by themselves be, like, an error?

cidc_api/shared/auth.py
jacoblurye: We were. Definitely open to reconsidering this going forward, though! Maybe we should throw a ticket on the backlog.

tests/resources/test_permissions.py
jacoblurye: Hmm, maybe - kind of a RBAC question, though, and right now that's enforced by the `/permissions` endpoint, which handles that issue by only allowing admins to grant permissions, rather than the `Permissions` model.

cidc_api/resources/downloadable_files.py
```diff
@@ -0,0 +1,134 @@
...
+
+    user = get_current_user()
+
+    # Ensure user has permission to access this file
+    if user.role != CIDCRole.ADMIN.value:
+        perm = Permissions.find_for_user_trial_type(
+            user.id, file_record.trial_id, file_record.upload_type
+        )
+        if not perm:
+            raise NotFound(f"No file with id {file_id}.")
```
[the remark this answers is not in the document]
jacoblurye: The issue is, there's no way to look up the appropriate permission for a user *without* an existing file record, since that file record is where we get the information about what upload type / trial id the permission needs to be for. Not sure there's an easy work around 🤷

PR #248, opened 2020-04-30
Improve (fix?) etag behavior

jacoblurye:
The way that #245 handled etag generation does not make sense. This PR fixes it by implementing the following behavior:
- only public columns are used to compute `_etag` values (columns that don't start with `_`)
- `_etag`s are only updated when the `CommonColumns.insert` and `CommonColumns.update` are used to modify a database record

### DIFF
--- a/tests/models/test_models.py
+++ b/tests/models/test_models.py
@@ -11,7 +11,6 @@ from sqlalchemy.orm.exc import NoResultFound
 
 from cidc_api.app import app
 from cidc_api.models import (
-    CommonColumns,
     Users,
     TrialMetadata,
     UploadJobs,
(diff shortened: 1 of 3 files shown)


tests/models/test_models.py
```diff
@@ -42,6 +42,24 @@ def wrapped(*args, **kwargs):
...
+    u._updated = datetime.now()
+    assert u.compute_etag() == etag
+
+    # Updates to public fields should change the etag
+    u.first_n = "foo"
+    new_etag = u.compute_etag()
+    assert new_etag != etag
+    u.first_n = "buzz"
+    assert u.compute_etag() != new_etag
+
```
curlup, line 61: nit: also check idempotence of  `compute_etag()`?

PR #249, opened 2020-05-01
Disable inactive users

jacoblurye:
Track system access on the `Users._accessed` field. Provide easy batch disabling of inactive users via the `Users.disable_inactive_users` method.

A daily cloud function cron job (not yet implemented) will do the actual disabling.


migrations/versions/922ee620c114_.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = "922ee620c114"
+down_revision = "37436458bc78"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
curlup, line 19: not sure why this is here.

jacoblurye: This is not intentional - I too am unsure. I'll delete the file.

PR #250, opened 2020-05-04
Add back in / route for CLI backwards compatibility

jacoblurye:


cidc_api/resources/__init__.py
```diff
@@ -19,3 +20,10 @@ def register_resources(app: Flask):
     app.register_blueprint(downloadable_files_bp, url_prefix="/downloadable_files")
     app.register_blueprint(upload_jobs_bp, url_prefix="/upload_jobs")
     app.register_blueprint(ingestion_bp, url_prefix="/ingestion")
+
+    # For backwards compatibility with the CLI, create an index route
+    # that checks the current user's auth status and nothing else
+    @app.route("/")
+    @requires_auth("index")
+    def index():
+        return "ok", 200
```
curlup, line 29: this line is not tested =)

jacoblurye: very true

PR #251, opened 2020-05-05
Fix exception handler to handle non-HTTPErrors and print tracebacks for internal server errors

jacoblurye:


cidc_api/app.py
```diff
@@ -32,31 +32,28 @@
...
 
-# JSON-ify HTTP errors
-@app.errorhandler(HTTPException)
-def jsonify_http_errors(e: HTTPException):
-    """Return JSON instead of default HTML for HTTP errors."""
-    data = {"code": e.code}
-    if hasattr(e, "exc") and isinstance(e.exc, ValidationError):
-        data["message"] = e.data["messages"]
+
+@app.errorhandler(Exception)
```
curlup, line 36: Not sure but maybe BaseException is the proper one?

jacoblurye: Flask only allows user-defined error handlers for subclasses of `Exception`.

PR #253, opened 2020-05-13
Trial metadata validation and RBAC updates

jacoblurye:
This PR:
* adds automatic cidc-schemas validation on the `TrialMetadata.metadata_json` field
* adds NCI biobank user permission on `/trial_metadata` resource endpoints

### DIFF
--- a/tests/resources/test_downloadable_files.py
+++ b/tests/resources/test_downloadable_files.py
@@ -1,7 +1,6 @@
 from datetime import datetime
 from typing import Tuple
 
-from cidc_schemas.prism import PROTOCOL_ID_FIELD_NAME
 from cidc_api.models import (
     Users,
     DownloadableFiles,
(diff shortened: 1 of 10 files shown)


tests/resources/test_downloadable_files.py
```diff
@@ -1,7 +1,6 @@
 from datetime import datetime
 from typing import Tuple
 
-from cidc_schemas.prism import PROTOCOL_ID_FIELD_NAME
```
curlup, line 4: Not sure about this change here.
Could you please elaborate on why you removed this import?

jacoblurye: It felt inconsistent to me to use this constant rather than just typing out the string `"protocol_identifier"`, since we don't have a similar one for any of the other fields in the metadata schema. The built-in metadata validation will now catch typos etc. for this and any other field.

PR #263, opened 2020-07-06
Fix permissions endpoint

jacoblurye:
This query param parsing bug currently prevents non-admin users from viewing their own permissions on staging and production (if you open the network tab in the developer console, you'll see 401 errors).

### DIFF
--- a/cidc_api/resources/permissions.py
+++ b/cidc_api/resources/permissions.py
@@ -28,7 +28,7 @@ permission_list_schema = PermissionListSchema()
 
 @permissions_bp.route("/", methods=["GET"])
 @requires_auth("permissions")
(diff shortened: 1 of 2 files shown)


cidc_api/resources/permissions.py
```diff
@@ -28,7 +28,7 @@
 
 @permissions_bp.route("/", methods=["GET"])
 @requires_auth("permissions")
-@use_args({"user_id": fields.Str()}, location="query")
+@use_args({"user_id": fields.Int()}, location="query")
```
jacoblurye, line 31: See line 50 below - the code was comparing a string to an integer, so `current_user.id != user_id` was always `True`.

PR #264, opened 2020-07-09
Add authentication via "upload tokens" for certain upload-related endpoints

jacoblurye:
This PR introduces "upload token authentication" for the `PATCH /upload_jobs` and `GET /ingestion/poll_upload_merge_status` endpoints via the `requires_upload_token_auth` decorator.

An upload token is a UUID randomly generated upon creation of a new `upload_job`. Possession of one of these tokens is considered sufficient to give a requesting user permission to update an upload job's status via `PATCH /upload_jobs` or to poll for an upload job's status via `GET /ingestion/poll_upload_merge_status`. This means that, even if a a user's identity token expires during a CLI upload, the CLI upload should still complete using this upload token.

**Note**: this change requires that we disable RBAC on endpoints wrapped with the `requires_upload_token_auth` decorator, since we are unable to determine a user's identity from an upload token.

This PR shouldn't be merged until the CLI has been updated appropriately.


cidc_api/resources/upload_jobs.py
```diff
@@ -86,18 +87,67 @@ def get_upload_job(upload_job: UploadJobs):
...
+            return endpoint(*pos_args, **kwargs)
+
+        return wrapped
+
+    return decorator
+
+
 @upload_jobs_bp.route("/<int:upload_job>", methods=["PATCH"])
-@requires_auth("upload_jobs", upload_job_roles)
+@requires_upload_token_auth(get_upload_id=lambda *a, **kw: kw["upload_job"])
```
curlup, line 143: nit: but then if this decorator is for upload job only and is very specific to this, why do this generic-ish `get_upload_id`.

So for me it's either should be more abstract or more upload_job specific (like specificities baked-in).

cidc_api/resources/upload_jobs.py
```diff
@@ -508,10 +554,7 @@ def poll_upload_merge_status():
     if not upload_id:
         raise BadRequest("Missing expected query parameter 'id'")
 
-    user = get_current_user()
-    upload = UploadJobs.find_by_id_and_email(upload_id, user.email)
-    if not upload:
-        raise NotFound(f"Could not find assay upload job with id {upload_id}")
+    upload = UploadJobs.find_by_id(upload_id)
```
curlup, line 557: nit: as additional sign that `UJ.find_by_id` should not be called within `requires_upload_token_auth` - it's duplicated here, which makes this controller to issue two identical db queries for no good reason.

cidc_api/models/models.py
```diff
@@ -561,6 +562,8 @@ class UploadJobs(CommonColumns):
     _status = Column(
         "status", Enum(*UPLOAD_STATUSES, name="upload_job_status"), nullable=False
     )
+    # A long, random identifier for this upload job
+    token = Column(String, default=uuid.uuid4)
```
curlup, line 566: will this propagate to the postgres' `default` on the column? 
Not sure it should, just trying to figure out if it's better to generate uuid in the app code or in the DB.

tests/resources/test_upload_jobs.py
```diff
@@ -172,6 +178,62 @@ def test_get_upload_job(cidc_api, clean_db, monkeypatch):
...
+    def endpoint(*args, **kwargs):
+        assert "upload_job" in kwargs
+        return "ok", 200
+
+    query_route = f"{test_route}/{job_id}"
+    nonexistent_job_id = "9999999"
+
+    # User must provide `token` query param
+    with cidc_api.test_request_context(query_route):
+        with pytest.raises(UnprocessableEntity):
```
curlup, line 200: why not add msg matching for the `.raises`? 
goes for every `pytest.raises` below as well.

cidc_api/models/models.py
jacoblurye: It won't - the default function runs app-side. I don't have a strong feeling on this, so I'm happy to go with whatever you think is best.

migrations/versions/4fc2addf0090_.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = "4fc2addf0090"
+down_revision = "37436458bc78"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("upload_jobs", sa.Column("token", sa.String(), nullable=True))
```
curlup, line 21: I wonder why string and not uuid type?

cidc_api/resources/upload_jobs.py
```diff
@@ -473,28 +550,20 @@ def upload_data_files(
...
-    [
-        CIDCRole.ADMIN.value,
-        CIDCRole.CIMAC_BIOFX_USER.value,
-        CIDCRole.CIDC_BIOFX_USER.value,
-    ],
-)
-def poll_upload_merge_status():
+@ingestion_bp.route("/poll_upload_merge_status/<int:upload_job>", methods=["GET"])
+@requires_upload_token_auth
+def poll_upload_merge_status(upload_job):
```
curlup, line 563:
```suggestion
def poll_upload_merge_status(upload_job: UploadJobs):
```
😎

cidc_api/resources/upload_jobs.py
```diff
@@ -86,18 +93,86 @@ def get_upload_job(upload_job: UploadJobs):
...
+        token = args["token"]
+        if str(token) != str(upload_job.token):
+            raise Unauthorized("upload_job token authentication failed")
+
+        # Pass the looked-up upload_job record to `endpoint` via the `upload_job` keyword argument
+        kwargs["upload_job"] = upload_job
+
+        return endpoint(*pos_args, **kwargs)
+
+    def _get_user():
```
curlup, line 132: nit: not sure why you made this and `_get_upload_job` as part of decorator definition and not module level declaration

cidc_api/resources/upload_jobs.py
```diff
@@ -86,18 +93,86 @@ def get_upload_job(upload_job: UploadJobs):
...
+    )
+
+    @wraps(endpoint)
+    @use_args(token_schema, location="query")
+    def wrapped(args, *pos_args, **kwargs):
+        # Attempt identity token authentication to get user info
+        user = _get_user()
+
+        # Try to find the upload_job associated with this request
+        upload_job = _get_upload_job(user, kwargs["upload_job"])
```
curlup, line 120: nit: Also I'm not sure for why inverse of control here - you provide `user` to this function only to check one condition,
which can be done without inverse 
```suggestion
        user = _get_user()

        # Try to find the upload_job associated with this request
        try:
            upload_job = _get_upload_job(kwargs["upload_job"])
        except (PreconditionRequired, NotFound) as e:
            # If there's an authenticated user associated with this request,
            # raise errors thrown by `lookup`. Otherwise, just report that auth failed.
            if user:
                raise e
            else:
                raise Unauthorized("upload_job token authentication failed")
```

cidc_api/models/models.py
curlup: I guess it won't work with postgres default - the value will only be available after an `insert` so no way to return it without either querying or sql `returning token`.
So I think app-side is fine.

migrations/versions/4fc2addf0090_.py
curlup: But at least I think this should not be nullable - as a precaution for app-side bug setting it empty for everything - thus basically losing auth.

PR #268, opened 2020-07-15
key setup for prism encrypt

curlup:


roshni-b:
Might want to bump API package version

PR #282, opened 2020-08-19
Add facet group column to DownloadableFiles and generate facet groups for existing file records

jacoblurye:
Existing file records `facet_group` column is set to their facet path in the current facet configuration, e.g., `Assay Type|mIF|Source Images`.

This PR only adds this migration. Updates to the faceted search implementation will come in a later PR.

### DIFF
--- a/migrations/versions/7590a700f9bc_.py
+++ b/migrations/versions/7590a700f9bc_.py
@@ -168,12 +168,15 @@ def build_facet_group_for_existing_files():
     """
     cases = []
     for path in list_facet_paths():
-        object_url_like = lambda exp: sa.text(f"object_url LIKE '{exp}'")
-        path_facets = sa.or_(*get_facets_for_path(object_url_like, path))
-        facet_group = "|".join(path)
-        cases.append((path_facets, facet_group))
+        path_facets = get_facets_for_path(lambda p: p, path)
+        for path_facet in path_facets:
+            clean_path_facet = path_facet.replace("%", "")
+            facet_group = "|".join([*path, clean_path_facet])
+            case_text = sa.text(f"object_url LIKE '{path_facet}'")
+            cases.append((case_text, facet_group))
 
-    return sa.case(cases)
+    # Files that don't currently have subfacets will default to [no facet group]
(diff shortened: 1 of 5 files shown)


migrations/versions/7590a700f9bc_.py
```diff
@@ -0,0 +1,201 @@
...
+
+    session = sa.orm.session.Session(bind=op.get_bind())
+    session.query(DownloadableFiles).update(
+        {"facet_group": facet_group_cases}, synchronize_session="fetch"
+    )
+    session.commit()
+
+    op.alter_column(
+        "downloadable_files", sa.Column("facet_group", sa.String(), nullable=False)
+    )
```
curlup, line 195: I think this should be one session. Maybe even with `add_column` but not sure.

migrations/versions/7590a700f9bc_.py
```diff
@@ -0,0 +1,201 @@
...
+    Builds a sqlalchemy `case` statement for figuring out which 
+    facet group a particular record belongs to.
+
+    This will build facet groups like "Assay Type|mIF|Source Images"
+    """
+    cases = []
+    for path in list_facet_paths():
+        object_url_like = lambda exp: sa.text(f"object_url LIKE '{exp}'")
+        path_facets = sa.or_(*get_facets_for_path(object_url_like, path))
+        facet_group = "|".join(path)
```
curlup, line 173: I think would be better to have more granular facet groups, eg `Assay|RNA|Source|rna_reads_.bam` etc.

migrations/versions/7590a700f9bc_.py
```diff
@@ -0,0 +1,201 @@
...
+facet_group_cases = build_facet_group_for_existing_files()
+
+
+def upgrade():
+    op.add_column(
+        "downloadable_files", sa.Column("facet_group", sa.String(), nullable=True)
+    )
+
+    session = sa.orm.session.Session(bind=op.get_bind())
+    session.query(DownloadableFiles).update(
```
curlup, line 188: I'd add an update for all the rest DFs to have non null facet group

PR #285, opened 2020-08-28
Batch download support

jacoblurye:
This PR does two things to advance us towards supporting batch downloads:
* introduces the `/downloadable_files/filelist` endpoint, which responds with a TSV mapping GCS URIs to local file paths
* adds functionality for granting and revoking conditional IAM policy bindings to individual users on certain trial/upload type combinations in the GCS data bucket, and ties this granting/revoking to insertion/deletion of `Permissions` records


cidc_api/models/models.py
```diff
@@ -330,6 +335,58 @@ class Permissions(CommonColumns):
...
+        grantee = Users.find_by_id(self.granted_to_user)
+        if grantee is None:
+            raise IntegrityError(
+                params=None,
+                statement=None,
+                orig=f"`granted_to_user` user must exist, but no user found with id {self.granted_to_user}",
+            )
+
+        # Grant IAM permission in GCS
+        grant_download_access(grantee.email, self.trial_id, self.upload_type)
```
curlup, line 355: I think this should go after `db.insert` so we won't have hangin IAM perms.

cidc_api/models/models.py
```diff
@@ -330,6 +335,58 @@ class Permissions(CommonColumns):
...
+        on the GCS data bucket.
+
+        NOTE: values provided to the `commit` argument will be ignored. This method always commits.
+        """
+        grantee = Users.find_by_id(self.granted_to_user)
+        if grantee is None:
+            raise NoResultFound(f"no user with id {self.granted_to_user}")
+
+        # Revoke IAM permission in GCS
+        revoke_download_access(grantee.email, self.trial_id, self.upload_type)
```
curlup, line 379: here it's in the right order. i.e. for failure scenario we at least won't have untracked IAM permissions.

cidc_api/models/models.py
curlup: Wait, how is this resolved? 

imagine session fails and rolls back, but `revoke_download_access` fails as well for whatever reason. In that scenario we end up with: IAM still granted but we don't have no corresponding record.

cidc_api/shared/gcloud_client.py
```diff
@@ -117,6 +118,111 @@ def revoke_upload_access(user_email: str):
...
+    member_id = f"user:{user_email}"
+    url_prefix, prefix_expression = _get_prefix_expression(trial_id, upload_type)
+    ttl_expression = _get_ttl_expression()
+
+    print(f"Granting download access on {url_prefix}* to {user_email}")
+
+    # get the current IAM policy for the data bucket
+    bucket = _get_bucket(GOOGLE_DATA_BUCKET)
+    policy = bucket.get_iam_policy(requested_policy_version=3)
+    policy.version = 3
```
curlup, line 133: not sure why that is needed? if it is - pls add a comment in the code.

jacoblurye: Will do - this is just following the example in the GCS docs, so I'll link to that.

PR #289, opened 2020-09-09
Consolidate IAM download permissions using '||'

jacoblurye:
Turns out GCP IAM only allows 20 bindings for the same member and role: https://cloud.google.com/iam/docs/conditions-overview

To avoid hitting that limit, this PR updates the way we manage conditional download bindings by `||`-chaining all of a user's trial / upload type permissions together in a single binding.


cidc_api/shared/gcloud_client.py
```diff
@@ -133,26 +135,25 @@ def grant_download_access(user_email: str, trial_id: str, upload_type: str):
...
-            }
+    # find the policy binding for this user if one exists
+    binding_index, binding = _get_user_download_binding(policy, user_email)
+
+    # add the permission clause in the binding condition expression if it isn't already included
+    expression = binding["condition"]["expression"]
+    if prefix_expression not in expression:
+        full_clause = f"({prefix_expression} && {ttl_expression})"
+        expression = (
+            f"{expression}{_DOWNLOAD_CLAUSE_OR}{full_clause}"
```
curlup, line 146: not sure but maybe like this? 
```suggestion
            f"({expression}){_DOWNLOAD_CLAUSE_OR}{full_clause}"
```

cidc_api/shared/gcloud_client.py
```diff
@@ -172,13 +172,51 @@ def revoke_download_access(user_email: str, trial_id: str, upload_type: str):
...
-    for binding in policy.bindings:
-        if not _is_matching_conditional_binding(binding, member_id, prefix_expression):
-            clean_bindings.append(binding)
+    # find the policy binding for this user if one exists
+    binding_index, binding = _get_user_download_binding(policy, user_email)
+
+    # remove the relevant clause from the condition if it exists
+    if binding_index is not None:
+        expression = binding["condition"]["expression"]
+        if prefix_expression in expression:
```
curlup, line 181: isn't that like an integrity error if prefix_expression *NOT* in expression? 
I suggest adding warning logging or something

cidc_api/shared/gcloud_client.py
```diff
@@ -172,13 +172,51 @@ def revoke_download_access(user_email: str, trial_id: str, upload_type: str):
...
+            filtered_clauses = [
+                clause
+                for clause in expression.split(_DOWNLOAD_CLAUSE_OR)
```
curlup, line 184: nit: I wonder if there's a python parser for CEL...

cidc_api/shared/gcloud_client.py
```diff
@@ -213,15 +251,42 @@ def _get_ttl_expression(days_until_expiry: int = INACTIVE_USER_DAYS) -> str:
...
-def _is_matching_conditional_binding(
-    binding: storage.bucket.Policy, member_id: str, subexpression: str
-) -> bool:
-    try:
-        has_user = binding["members"] == {member_id}
-        has_condition = subexpression in binding["condition"]["expression"]
-        return has_user and has_condition
-    except KeyError:
-        return False
+def _get_user_download_binding(
```
curlup, line 254: nit: rather then returning an index of a binding (later used to either overwrite or to delete the binding by index)
this fn could been something like `find_and_pop_user_download_binding`, so it will already remove the binding from the policy and overwrite would be as easy as appending new, proper binding to the policy. I can't shake off a (somewhat groundless) feeling that substitution by index is fragile or something.

cidc_api/shared/gcloud_client.py
```diff
@@ -213,15 +251,42 @@ def _get_ttl_expression(days_until_expiry: int = INACTIVE_USER_DAYS) -> str:
...
+    # try to find the policy binding on the `policy`
+    user_binding = None
+    user_binding_index = None
+    for i, binding in enumerate(policy.bindings):
+        try:
+            user_is_member = binding["members"] == {member_id}
+            role_is_download = binding["role"] == GOOGLE_DOWNLOAD_ROLE
+            if user_is_member and role_is_download:
+                user_binding = binding
+                user_binding_index = i
```
curlup, line 273: Maybe
```suggestion
                user_binding_index = i
                break
```
? 

Which brings a question of why do we return only one (first, last) binding if user might have multiple?
Shouldn't it raise a flag or try to join multiple together or something?

cidc_api/shared/gcloud_client.py
```diff
@@ -133,26 +135,25 @@ def grant_download_access(user_email: str, trial_id: str, upload_type: str):
...
+    # find the policy binding for this user if one exists
+    binding_index, binding = _get_user_download_binding(policy, user_email)
+
+    # add the permission clause in the binding condition expression if it isn't already included
+    expression = binding["condition"]["expression"]
+    if prefix_expression not in expression:
+        full_clause = f"({prefix_expression} && {ttl_expression})"
+        expression = (
+            f"{expression}{_DOWNLOAD_CLAUSE_OR}{full_clause}"
+            if expression
```
curlup, line 147: I'd add comment here that for newly created binding the expression will be empty.

cidc_api/shared/gcloud_client.py
curlup: This would also allow to split this function in two - `pop` and `create_empty` - and to call the later in lazy way:
```
   binding =  find_and_pop_user_download_binding() or create_empty_download_binding()
```

curlup (commented):
Small things, overall looks nice!

cidc_api/shared/gcloud_client.py
jacoblurye: Sure, a warning sounds good.

cidc_api/shared/gcloud_client.py
jacoblurye: After migrating from the previous implementation of IAM download bindings, a user should have at most one binding that matches this condition. Should the code raise a warning or error if this invariant is violated?

curlup: yeah, maybe a warning in the logs for us to catch?

migrations/versions/15402497cb30_add_iam_download_permissions.py
```diff
@@ -8,7 +8,7 @@
 from alembic import op
 import sqlalchemy as sa
 
-from cidc_api.models import Permissions
+# from cidc_api.models import Permissions
```
curlup, line 11: hm, and all commentings out below in this file

jacoblurye: I can rollback the prod + staging dbs and delete the migration, if that's preferable.

curlup: Not sure what you mean, could you pls provide context? I'm a bit lost - rollback ? delete which\ migration?

curlup (approved):
though I'm not sure what's happening with all that migration stuff

migrations/versions/15402497cb30_add_iam_download_permissions.py
jacoblurye: The file you've commented on here is a migration intended to automatically add IAM download permissions based on the records in the `permissions` table (this probably should've just been a one-off script one of us runs rather than an automatic migration, which is what we've decided to do to integrate updates from this PR). This migration has already been applied on both the staging and production dbs, so Alembic, the migration version manager we use, will encounter errors when next running migrations if this file is simply deleted. Rolling back (or `downgrade`-ing in Alembic terms) to the previous migration version in both environments will make it safe to delete this file, though.

PR #298, opened 2020-09-17
Add migration to consolidate facet groups

jacoblurye:


curlup (approved):
Oh, me likey very much!

PR #300, opened 2020-09-22
Trial-file bundles

jacoblurye:
Add option to the `/trial_metadata` endpoint to list trials with their "file bundles": a collection of file IDs associated with that trial, organized by upload type and file purpose. These file bundles are intended primarily for building download batches in trial overview cards in the portal.


cidc_api/models/models.py
```diff
@@ -456,6 +464,38 @@ class ValidationMultiError(Exception):
...
+        <type 1>: {
+            <purpose 1>: [<file id 1>, <file id 2>, ...],
+            <purpose 2>: [...]
+        },
+        <type 2>: {...}
+    }
+    ```
+    """
+    file_bundles: FileBundle = {}
+    for f in files:
```
curlup, line 488: itertools.groupby?

cidc_api/models/models.py
```diff
@@ -606,6 +646,76 @@ def generate_sample_csv(cls, trial_id: str, session: Session) -> str:
             raise NoResultFound(f"No trial found with id {trial_id}")
         return unprism.unprism_samples(trial.metadata_json)
 
+    file_bundle: Optional[FileBundle]
+
+    @classmethod
+    @with_default_session
+    def list_with_file_bundles(cls, session: Session, **pagination_args):
+        """TODO: docstring"""
```
curlup, line 654: :)

cidc_api/models/models.py
```diff
@@ -606,6 +646,76 @@ def generate_sample_csv(cls, trial_id: str, session: Session) -> str:
...
+                        DownloadableFiles.consolidated_upload_type,
+                        "purpose",
+                        DownloadableFiles.file_purpose,
+                        "id",
+                        DownloadableFiles.id,
+                    )
+                ),
+            )
+            .filter(TrialMetadata.trial_id == DownloadableFiles.trial_id)
+            .group_by(TrialMetadata)
```
curlup, line 677: if you either way doing group_by here,
why not make it to final view (the one that `_build_file_bundle` makes now) right here with SQL group by `type` and `purpose` ?

cidc_api/models/models.py
```diff
@@ -606,6 +646,76 @@ def generate_sample_csv(cls, trial_id: str, session: Session) -> str:
...
+        )
+        # The above query excludes trials with no files, so add them here
```
curlup, line 679: Do we need trials with no files?

Have you checked how this (afaik some sort of fancy left outer join or something) will actually be executed? 
It's just that maybe if we don't quite need trials with no files I would for sure prefer this to be simpler, as now I'm having hard time figuring out what's happening here, so Im concerned.

cidc_api/resources/trial_metadata.py
```diff
@@ -25,13 +26,28 @@
...
-@requires_auth("trial_metadata", trial_modifier_roles)
-@use_args_with_pagination({}, trial_metadata_schema)
+@requires_auth("trial_metadata")
+@use_args_with_pagination(trial_filter_schema, trial_metadata_schema)
 @marshal_response(trial_metadata_list_schema)
 def list_trial_metadata(args, pagination_args):
     """List all trial metadata records."""
-    trials = TrialMetadata.list(**pagination_args)
+    user = get_current_user()
+    include_file_bundles = bundle_argname in args and args.pop(bundle_argname)
```
curlup, line 43: nit: just `.pop` with default?

curlup (commented):
stuff to discuss 
* additional `group by`s moved to sql 
* fancy sql on getting trials without files. Can we maybe just have another query just for that and then concat results?

cidc_api/resources/trial_metadata.py
jacoblurye: TIL :)

tests/models/files/test_details.py
```diff
@@ -0,0 +1 @@
+# No tests here...yet.
```
jacoblurye, line 1: Don't think there are any appropriate unit tests for this module, since right now it's basically just a big dictionary.

cidc_api/resources/downloadable_files.py
```diff
@@ -80,27 +80,26 @@ def get_filelist(args):
...
-    batch_id_filter = lambda q: user_perms_filter(
-        q.filter(DownloadableFiles.id.in_(file_ids))
-    )
 
-    # Get requested files
-    files = DownloadableFiles.list(filter_=batch_id_filter)
+    # Get request object_urls
+    urls = DownloadableFiles.list_object_urls(
+        args["file_ids"], filter_=user_perms_filter
+    )
```
jacoblurye, line 90: This fixes a relevant bug with the `/downloadable_files/filelist` endpoint, where filelist files were getting truncated by the default pagination functionality of `DownloadableFiles.list`.

cidc_api/models/models.py
```diff
@@ -1025,9 +1091,92 @@ def get_by_object_url(object_url: str, session: Session):
...
+                    func.json_agg(cls.id).label("ids"),
+                ]
+            )
+            .group_by(cls.trial_id, cls.consolidated_upload_type, cls.file_purpose)
+            .alias("id_bundles")
+        )
+        purpose_bundles = (
+            select(
+                [
+                    literal_column("id_bundles.trial_id"),
```
curlup, line 1136: nit: wouldn't it work with just `cls.trial_id`? I meant in psql it worked without subquery alias prefix (`id_bundles.`

cidc_api/models/models.py
```diff
@@ -1025,9 +1091,92 @@ def get_by_object_url(object_url: str, session: Session):
...
+                    literal_column("id_bundles.type"),
+                    func.json_object_agg(
+                        func.coalesce(
+                            literal_column("id_bundles.purpose"), "miscellaneous"
```
curlup, line 1140: I think we have something like `[no purpose]` as misc option already (in db records?)

curlup (approved):

cidc_api/models/models.py
jacoblurye: It wouldn't - doing so was what produced a `SQL` query with repeated self-joins. However, removing the prefixes and just using `literal_column("trial_id")` etc. does work, so I'll update the code accordingly.

cidc_api/models/models.py
jacoblurye: We don't actually. We have a value like `[no facet group]` in certain file records `facet_group` field, but file purposes are a) are purely views not materialized in the database and b) only have four "allowed" values (not strictly enforced), `analysis`, `source`, `clinical`, and `miscellaneous`, based on this https://github.com/CIMAC-CIDC/cidc-ngs-pipeline-api/blob/2c7887a0a312dd4ec87fbe85a2741242977b38ea/output_API.schema.json#L28

PR #306, opened 2020-10-06
Migrate existing IHC files to use more specific facet groups

jacoblurye:
The IHC analogue to #302.

### DIFF
--- a/cidc_api/models/files/facets.py
+++ b/cidc_api/models/files/facets.py
@@ -197,9 +197,8 @@ assay_facets: Facets = {
         )
     },
     "IHC": {
-        "All IHC Files": FacetConfig(
-            ["Assay Type|IHC|All IHC Files|/ihc", "/ihc/ihc_image."]
-        )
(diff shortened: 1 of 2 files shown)


cidc_api/models/files/facets.py
```diff
@@ -197,9 +197,8 @@ def __init__(self, match_clauses: List[str], description: Optional[str] = None):
         )
     },
     "IHC": {
-        "All IHC Files": FacetConfig(
-            ["Assay Type|IHC|All IHC Files|/ihc", "/ihc/ihc_image."]
-        )
+        "Images": FacetConfig(["/ihc/ihc_image."]),
```
curlup, line 200: dot at the end feels weird, but what do i know

curlup (approved):

cidc_api/models/files/facets.py
jacoblurye: That's the auto-generated facet group from prism.

PR #309, opened 2020-10-08
Add endpoint for getting 'related' files 

jacoblurye:
Use a file's `additional_metadata` and `data_category` values to determine a list of files that are meaningfully related to that file.

PR is a draft because I need to more thoroughly test the `DownloadableFiles.get_related_files` method.

### DIFF
--- a/cidc_api/models/models.py
+++ b/cidc_api/models/models.py
@@ -994,9 +994,11 @@ class DownloadableFiles(CommonColumns):
               has an associated `cimac_id`.
             * isn't sample-specific, and relates to the same `data_category_prefix`.
         """
+        # If this file has an associated sample, get other files associated with that sample.
+        # Otherwise, get other non-sample-specific files for this trial and data category.
         if self.cimac_id is not None:
             query = text(
-                "SELECT downloadable_files.* "
+                "SELECT DISTINCT downloadable_files.* "
                 "FROM downloadable_files, LATERAL jsonb_each_text(additional_metadata) addm_kv "
                 "WHERE addm_kv.value LIKE :cimac_id AND trial_id = :trial_id AND id != :id"
             )
@@ -1009,8 +1011,6 @@ class DownloadableFiles(CommonColumns):
                 session.execute(query, params), DownloadableFiles
             )
         else:
-            # If a file's additional metadata JSON has no key that ends with `.cimac_id`,
-            # then we consider that file "not sample specific".
             not_sample_specific = not_(
                 literal_column("additional_metadata::text").like('%.cimac_id":%')
             )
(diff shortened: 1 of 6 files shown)


cidc_api/models/models.py
```diff
@@ -970,6 +972,61 @@ def file_purpose(self):
...
+    def cimac_id(self):
+        """
+        Extract the `cimac_id` associated with this file, if any, by searching the file's 
+        additional metadata for a field with a key like `<some>.<path>.cimac_id`.
+
+        NOTE: this is not a sqlalchemy hybrid_property, and it can't be used directly in queries.
+        """
+        for key, value in self.additional_metadata.items():
+            if key.endswith("cimac_id"):
+                return value
```
curlup, line 985: this shows up as uncovered

cidc_api/models/models.py
```diff
@@ -970,6 +972,61 @@ def file_purpose(self):
...
+        Return a list of file records related to this file. We could define "related"
+        in any number of ways, but currently, a related file:
+            * is sample-specific, and relates to the same sample as this file if this file 
+              has an associated `cimac_id`.
+            * isn't sample-specific, and relates to the same `data_category_prefix`.
+        """
+        if self.cimac_id is not None:
+            query = text(
+                "SELECT downloadable_files.* "
+                "FROM downloadable_files, LATERAL jsonb_each_text(additional_metadata) addm_kv "
```
curlup, line 1000: I think this could potentially return multiple instances of the same DF record, in a case where `additional_metadata={"something.cimac_id": "Csomecimac.id", "something_else.cimac_id": "Csomecimac.id"}`
I'm mentioning this as I'm not sure if client code will be prepared to that.

cidc_api/models/models.py
```diff
@@ -970,6 +972,61 @@ def file_purpose(self):
...
+            )
+            params = {
+                "cimac_id": f"%{self.cimac_id}",
+                "trial_id": self.trial_id,
+                "id": self.id,
+            }
+            related_files = result_proxy_to_models(
+                session.execute(query, params), DownloadableFiles
+            )
+        else:
```
curlup, line 1011: oh, I though that we might show not_sample_specific files always, not only when there's no cimac_id related ones.

cidc_api/models/models.py
```diff
@@ -1202,3 +1259,15 @@ def build_file_bundle_query(cls) -> Query:
...
+
+def result_proxy_to_models(
+    result_proxy: ResultProxy, model: BaseModel
+) -> List[BaseModel]:
+    model_instances = []
+    for row_proxy in result_proxy:
+        row_dict = {}
+        for key, value in row_proxy.items():
+            row_dict[key] = value
+        model_instances.append(model(**row_dict))
```
curlup, line 1272:
```suggestion
        row_dict = dict(row_proxy.items())
        model_instances.append(model(**row_dict))
```

curlup: or even

```suggestion
        model_instances.append(model(**dict(row_proxy.items())))
```

curlup (approved):
looks good

cidc_api/models/models.py
jacoblurye: see draft description 😄

cidc_api/models/models.py
jacoblurye: Would this be resolved by doing `SELECT DISTINCT downloadable_files.* ...` instead? I'll ensure the tests cover this case, either way, since it's a sneaky one.

cidc_api/models/models.py
jacoblurye: Certainly wouldn't hurt! I'm happy to update the behavior accordingly.

cidc_api/models/models.py
jacoblurye: Oh, very nice!

PR #311, opened 2020-10-08
Added logging for writing/deleting to the permissions table

scvannost:
new deleted_by param on Permissions.delete
throw error if no deleted_by given
update tests to check for error if don't give delete_by
update api calls to provide deleted_by


curlup (approved):
I think it looks good, but can't shake off the feeling I'm missing something.
@jacoblurye can you pls take a look as well

jacoblurye (commented):
I'd suggest testing that the log statements appear when you'd expect using pytest's `capsys` fixture: https://docs.pytest.org/en/stable/capture.html#accessing-captured-output-from-a-test-function

PR #318, opened 2020-10-21
More robust default value behavior for additional metadata

jacoblurye:
Previously, only `DownloadableFiles.create_from_metadata` was checking for null-ish `additional_metadata` values, which meant that it was still possible to set `additional_metadata` to `"null"` (JSON null, not SQL null) via, e.g., `DownloadableFiles.create_from_blob`.

This lead to derivative files getting created with null `additional_metadata` here: https://github.com/CIMAC-CIDC/cidc-cloud-functions/blob/d9339a301c6d473bbf7331cb07c1d3458a2ac598/functions/upload_postprocessing.py#L99

This PR adds a sqlalchemy validator on the `DownloadableFiles.additional_metadata` attribute that encodes appropriate defaulting behavior for all `__setattr__` operations on that attribute.

### DIFF
--- a/cidc_api/models/models.py
+++ b/cidc_api/models/models.py
@@ -1019,6 +1019,10 @@ class DownloadableFiles(CommonColumns):
                 return value
         return None
 
+    @validates("additional_metadata")
+    def check_additional_metadata_default(self, key, value):
+        return {} if value in ["null", None, {}] else value
+
     @with_default_session
     def get_related_files(self, session: Session) -> list:
         """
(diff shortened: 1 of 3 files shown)


curlup (approved):
no migration tho, not needed or do you plan to it manually?

jacoblurye:
Instead of a migration, I'm planning to manually retrigger the `upload_postprocessing` cloud function, because right now the only files with `additional_metadata = 'null'` were generated by that function.

PR #325, opened 2020-10-28
V0 21 1 to v0 22 0

curlup:


scvannost (approved):

migrations/versions/c121b5fc5973_.py
```diff
@@ -0,0 +1,28 @@
+"""empty message
```
jacoblurye, line 1: I didn't know about this until recently, but you can add a descriptive messages to new migrations, like:
```bash
FLASK_APP=cidc_api.app:app flask db migrate -m "encrypt participant identifiers"
```
Most usefully, this gives the file a more meaningful name, like `<hash>_encrypt_participant_identifiers.py`.

jacoblurye (approved):

PR #329, opened 2020-10-29
Add /info/data_overview endpoint

jacoblurye:
This PR adds a public endpoint that produces simple stats about the data currently in the system.


scvannost (approved):

curlup (approved):
seems good, but with this the DM leaks into the API repo. Not sure how bad it is, but would love to discuss

PR #331, opened 2020-11-05
Add back code for sending email notification to newly-approved users

jacoblurye:
I think I accidentally deleted this functionality during #245's major refactor. A cautionary tale about having good test coverage before refactoring code!

### DIFF
--- a/.env
+++ b/.env
@@ -9,7 +9,7 @@ AUTH0_CLIENT_ID='Yjlt8LT5vXFJw1Z8m8eaB5aZO26uPyeD'
 PRISM_ENCRYPT_KEY='key'
 
 ENV='dev'
-DEBUG=True
+DEBUG=False
 TEST_POSTGRES_URI = 'postgresql://cidcdev:1234@localhost:5432/cidctest'
 
 # Uncomment to use local cloud functions
--- a/cidc_api/models/models.py
+++ b/cidc_api/models/models.py
@@ -65,6 +65,7 @@ from ..shared.gcloud_client import (
     publish_upload_success,
     grant_download_access,
     revoke_download_access,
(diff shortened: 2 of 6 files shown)


cidc_api/models/models.py
```diff
@@ -252,6 +253,14 @@ class Users(CommonColumns):
     role = Column(Enum(*ROLES, name="role"))
     disabled = Column(Boolean, default=False, server_default="false")
 
+    @validates("approval_date")
+    def send_approval_confirmation(self, key, new_approval_date):
+        """Send this user an approval email if their account has just been approved"""
+        if self.approval_date is None and new_approval_date is not None:
```
scvannost, line 259: what should happen if `self.approval_date is None`? especially if `new_approval_date is not None`

scvannost (approved):

.env
```diff
@@ -9,7 +9,7 @@ AUTH0_CLIENT_ID='Yjlt8LT5vXFJw1Z8m8eaB5aZO26uPyeD'
 PRISM_ENCRYPT_KEY='key'
 
 ENV='dev'
-DEBUG=True
+DEBUG=False
```
curlup, line 12: ? this is dev right?

curlup (approved):

.env
jacoblurye: Oh, whoops! Left over from local testing - thanks.

cidc_api/models/models.py
jacoblurye: It should send the confirmation email in that case (not sure I understand the question?).

scvannost: `self._approval_date is not None`, sorry

PR #340, opened 2020-11-23
Don't log with print statements

jacoblurye:
Instead,
* use Flask's built in logger
* update gunicorn's logging config
* update tests that read from stdout

### DIFF
--- a/gunicorn.conf.py
+++ b/gunicorn.conf.py
@@ -1,6 +1,10 @@
 """Configuration for the gunicorn WSGI server."""
 import os
 
-bind = f':{os.environ.get("PORT", 8080)}'
-loglevel = "DEBUG"
+is_dev = os.environ.get("ENV") == "dev"
+
+# See https://docs.gunicorn.org/en/stable/settings.html
+port = os.environ.get("PORT", 8080)
+loglevel = "DEBUG" if is_dev else "INFO"
+reload = is_dev
 timeout = 60
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ If you're updating `models.py`, you should create a migration and commit the res
 
 Once you have a development database set up and running, run the API server:
 ```bash
-python cidc_api/app.py
+ENV=dev gunicorn cidc_api.app:app
 ```
 
 ## Testing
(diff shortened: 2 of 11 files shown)


gunicorn.conf.py
```diff
@@ -1,6 +1,10 @@
...
 import os
 
-bind = f':{os.environ.get("PORT", 8080)}'
-loglevel = "DEBUG"
+is_dev = os.environ.get("ENV") == "dev"
+
+# See https://docs.gunicorn.org/en/stable/settings.html
+port = os.environ.get("PORT", 8080)
+loglevel = "DEBUG" if is_dev else "INFO"
+reload = is_dev
```
jacoblurye, line 9: I didn't realize until now that `gunicorn` could reload-on-code-change, but it can!

curlup (changes requested):
as I'm in general all for logger instead of printing, I'm just wondering if you really think this might be a perf bottle neck (I would be surprised).

Also I think all those `logger().info(f"\t{e}")` are a bit two verbose in two ways:
* usually one would get a logger instance on the module level and use it as `logger.info(...)`. AFAIK there's no need for an additional call each time to dynamically resolve/get the actual logger. also current name `logger` is a bit misleading then - it's a loggerFactory or something, if you need to call it to get a logger. The usual way as I know (but please also check with `logging howto` or `cookbook` in python's docs) is just do 
```
import logging
logger = logging.getLogger("my.custom.module.name" or __name__)
```
in each module that's it. Is there something i'm missing with regard to Flask here?

* `logger.info(f"\t{e}")` - I don't think there's much sense in `\t` now, as this was poor mans formatting. I would just use `logger.error(e)` or `logger.info(f"validation: {e}")` or something if you want to add context.

jacoblurye:
I *think* there may be something you're missing with regard to Flask - in particular, that the "proper" way to log in a Flask app is to use `current_app.logger`, which is what the admittedly poorly named `logger()` factory function provides if an `app_context` is available. An `app_context` will be available whenever the API is actually running - meaning `logger()` will *always* return `current_app.logger` in this case - but if no `app_context` is available, `current_app` won't be defined, and this will lead to a crash. I agree that dynamic solution feels not ideal, and it's worth my taking a second pass at this in another PR to consider alternatives.

Those `\t`s etc. are a relic of the fact that I basically did a find-and-replace here to get rid of print statements without actually changing their content. It's definitely worth making a second pass through to reformat and choose appropriate log levels for existing print statements.

PR #356, opened 2020-12-04
filelist endpoint: accept ids via JSON body

jacoblurye:
Passing file IDs via the URL query string breaks for large numbers of files, since the maximum URL length gets exceeded, so accept the `file_ids` list via a JSON request body instead.

This is a breaking change, and a UI update will need to accompany it.

### DIFF
--- a/cidc_api/resources/downloadable_files.py
+++ b/cidc_api/resources/downloadable_files.py
@@ -86,12 +86,10 @@ def get_related_files(downloadable_file: DownloadableFiles):
     return {"_items": downloadable_file.get_related_files()}
 
 
-@downloadable_files_bp.route("/filelist", methods=["GET"])
+@downloadable_files_bp.route("/filelist", methods=["POST"])
 @requires_auth("filelist")
-@use_args(
-    {"file_ids": fields.DelimitedList(fields.Int, required=True)}, location="query"
-)
-def get_filelist(args):
+@use_args({"file_ids": fields.List(fields.Int, required=True)}, location="json")
+def generate_filelist(args):
     """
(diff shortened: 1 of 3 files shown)


cidc_api/resources/downloadable_files.py
```diff
@@ -86,12 +86,10 @@ def get_related_files(downloadable_file: DownloadableFiles):
     return {"_items": downloadable_file.get_related_files()}
 
 
-@downloadable_files_bp.route("/filelist", methods=["GET"])
+@downloadable_files_bp.route("/filelist", methods=["POST"])
```
curlup, line 89: just curious why not keep both and not make it a breaking change?

curlup (approved):
LGTM

just curious again if this is for pre-bundled files or for ad-hoc file sets that get too large?
if it's the former, i'd consider making a different endpoint that wouldn't require explicit list of ids of all the files in the bundle X (say WES) for trial Y, but will only require X and Y as inputs. Let me know if it's hard to understand what I mean.

cidc_api/resources/downloadable_files.py
jacoblurye: My thinking is, the current code is already broken for large file batches, so a breaking change doesn't hurt *that much* more 🤷

It's true that keeping both wouldn't really hurt either, aside from potentially causing confusion.

jacoblurye:
This endpoint is for all file bundles, both "pre-bundled" and ad-hoc. I see how having a separate pre-bundled filelist endpoint with a simpler interface could allow for cleaner requests, but I guess I'm not totally clear on the advantage beyond that.

scvannost (approved):

PR #359, opened 2020-12-08
Change API representation of DownloadableFiles to also carry descriptions

scvannost:


cidc_api/models/models.py
```diff
@@ -1067,6 +1067,14 @@ def data_category_prefix(cls):
     def file_purpose(self):
         return details_dict.get(self.facet_group).file_purpose
 
+    @hybrid_property
+    def short_description(self):
+        return details_dict.get(self.facet_group).short_description
+
+    @hybrid_property
+    def long_description(self):
+        return details_dict.get(self.facet_group).long_description
```
jacoblurye, line 1076: My mistake, I misled you here - since these values won't be involved in sqlalchemy queries, we actually don't want them to be `@hybrid_property`s. They should just be plain`@property`s, like `cimac_id` is below.

Also, I'd prefer for this code to go below the `@file_purpose.expression` definition, since it's better not to visually separate the two `file_purpose`-related methods.

jacoblurye (changes requested):

jacoblurye (approved):

PR #369, opened 2020-12-21
Refresh download permissions

jacoblurye:
* Add a new method `Permissions.refresh_iam_permissions` that extends the TTLs on a given user's GCS download permissions. 
* Refresh a user's GCS download permissions when an admin reactivates their account via the `/users` endpoint.

Up next: a daily cloud function that refreshes active users' download permissions to keep their IAM permissions (more) in-sync with their account status.

### DIFF
--- a/cidc_api/models/models.py
+++ b/cidc_api/models/models.py
@@ -493,10 +493,10 @@ class Permissions(CommonColumns):
 
     @staticmethod
     @with_default_session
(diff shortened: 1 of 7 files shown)


curlup (commented):
looks good, needs  `refresh_iam_permissions` coverage

curlup (approved):

PR #370, opened 2020-12-28
Intake bucket IAM permission management

jacoblurye:
This PR introduces functionality related granting users access to sub-buckets within the new data "intake" bucket. It adds an `/ingestion/intake_gcs_uris` endpoint for granting new permissions and listing currently granted permissions and various functions in `gcloud_client` related to intake bucket IAM permissions management.


cidc_api/config/settings.py
```diff
@@ -55,15 +55,16 @@
 ALGORITHMS = ["RS256"]
 
 ### Configure GCP ###
-GOOGLE_CLOUD_PROJECT = environ.get("GOOGLE_CLOUD_PROJECT")
-GOOGLE_UPLOAD_BUCKET = environ.get("GOOGLE_UPLOAD_BUCKET")
-GOOGLE_UPLOAD_TOPIC = environ.get("GOOGLE_UPLOAD_TOPIC")
-GOOGLE_DATA_BUCKET = environ.get("GOOGLE_DATA_BUCKET")
-GOOGLE_UPLOAD_ROLE = environ.get("GOOGLE_UPLOAD_ROLE")
+GOOGLE_CLOUD_PROJECT = environ["GOOGLE_CLOUD_PROJECT"]
```
jacoblurye, line 58: We want to get `KeyError`s if these environment variables aren't set, because they're critical to the API functioning correctly.

cidc_api/shared/gcloud_client.py
```diff
@@ -121,60 +122,174 @@ def revoke_upload_access(user_email: str):
...
+    bucket = _get_bucket(GOOGLE_INTAKE_BUCKET)
+    policy = bucket.get_iam_policy(requested_policy_version=3)
+    policy.version = 3
+
+    user_uris = []
+    for binding in policy.bindings:
+        expression = binding.get("condition", {}).get("expression", "")
+        subdir_match = intake_subdir_regex.match(expression)
+        if (
+            binding["members"] == {user_member(user_email)}
```
scvannost, line 190: should this be `user_member(user_email) in binding["members"]`?

cidc_api/shared/gcloud_client.py
```diff
@@ -193,68 +308,64 @@ def revoke_all_download_access(user_email: str):
...
     # try to find the policy binding on the `policy`
     user_binding_index = None
     for i, binding in enumerate(policy.bindings):
-        user_is_member = binding.get("members") == {member_id}
-        role_is_download = binding.get("role") == GOOGLE_DOWNLOAD_ROLE
-        has_prefix = prefix_clause in binding.get("condition", {}).get("expression", "")
-        if user_is_member and role_is_download and has_prefix:
+        if (
+            binding.get("role") == role
+            and binding.get("members") == {user_member(user_email)}
```
scvannost, line 361: similarly, should this be `user_member(user_email) in binding.get("members")`

cidc_api/shared/gcloud_client.py
jacoblurye: So, we're actually looking specifically for bindings that contain *only* this user as a member - if that weren't the case, then `in` would be more appropriate. I'll add a comment to this effect.

cidc_api/resources/upload_jobs.py
```diff
@@ -618,3 +618,43 @@ def extra_assay_metadata():
...
+@ingestion_bp.route("/intake_gcs_uri", methods=["POST"])
+@requires_auth("intake_gcs_uri", INTAKE_ROLES)
+@use_args(
+    {"trial_id": fields.Str(required=True), "upload_type": fields.Str(required=True)}
+)
+def create_intake_gcs_uri(args):
+    """
+    Grant upload access to the current user for the provided trial and upload types,
+    returning the GCS URI the user now has permission to upload to.
+    NOTE: we don't verify that the user has permission to upload data for this trial-upload
```
curlup, line 639: still feels weird.

curlup (approved):

PR #373, opened 2021-01-14
Add handling for missing files in process_as prefix

scvannost:
Schemas: https://github.com/CIMAC-CIDC/cidc-schemas/pull/406
CLI: https://github.com/CIMAC-CIDC/cidc-cli/pull/135
API: https://github.com/CIMAC-CIDC/cidc-api-gae/pull/373

### DIFF
--- a/cidc_api/resources/upload_jobs.py
+++ b/cidc_api/resources/upload_jobs.py
@@ -145,7 +145,9 @@ def requires_upload_token_auth(endpoint):
 @upload_jobs_bp.route("/<int:upload_job>", methods=["PATCH"])
 @requires_upload_token_auth
 @unmarshal_request(
-    UploadJobSchema(only=["status", "token"]), "upload_job_updates", load_sqla=False
+    UploadJobSchema(only=["status", "gcs_file_map", "token"]),
+    "upload_job_updates",
+    load_sqla=False,
 )
 @marshal_response(upload_job_schema, 200)
 def update_upload_job(upload_job: UploadJobs, upload_job_updates: UploadJobs):
@@ -497,6 +499,7 @@ def upload_data_files(
     upload_moment = datetime.datetime.now().isoformat()
     uri2uuid = {}
     url_mapping = {}
+    optional_files = []
     files_with_extra_md = {}
     for file_info in file_infos:
         uuid = file_info.upload_placeholder
@@ -517,6 +520,9 @@ def upload_data_files(
         if file_info.metadata_availability:
             files_with_extra_md[file_info.local_path] = file_info.upload_placeholder
 
+        if file_info.allow_empty:
+            optional_files.append(file_info.local_path)
+
     gcs_blob = gcloud_client.upload_xlsx_to_gcs(
         trial.trial_id, "assays", template_type, xlsx_file, upload_moment
     )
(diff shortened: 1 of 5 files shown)


jacoblurye (commented):
This looks good, but I'd like to see additional tests showing that the new `optional_files` array gets built as expected and that the `gcs_file_map` can be updated via the `PATCH /upload_jobs` endpoint.

jacoblurye:
Oh, also - be sure to update `MIN_CLI_VERSION` https://github.com/CIMAC-CIDC/cidc-api-gae/blob/79be13bf7b26e699ff09b7421e0ff1f06eed97a2/cidc_api/config/settings.py#L33 to the new CLI version in your other PR https://github.com/CIMAC-CIDC/cidc-cli/pull/135/files#diff-14c6359e27f50470f636a43aeafed472f9c51b203b5ee84e1384086c2bef9472R1

cidc_api/resources/upload_jobs.py
```diff
@@ -145,12 +145,24 @@ def wrapped(args, *pos_args, **kwargs):
...
 )
 @marshal_response(upload_job_schema, 200)
-def update_upload_job(upload_job: UploadJobs, upload_job_updates: UploadJobs):
+def update_upload_job(upload_job: UploadJobs, upload_job_updates: dict):
     """Update an upload_job."""
     try:
+        if "gcs_file_map" in upload_job_updates and upload_job.gcs_file_map is not None:
+            upload_job_updates["metadata_patch"] = upload_job.metadata_patch.copy()
+            for uri, uuid in upload_job.gcs_file_map.items():
+                if uri not in upload_job_updates["gcs_file_map"]:
```
jacoblurye, line 159: Do we ever expect a large number of missing optional files? If so, this will run into the same performance issue for merging many artifacts that was solved with the `uuid_path_map` optimization for batch artifact merging here https://github.com/CIMAC-CIDC/cidc-schemas/blob/b74fb7944ef25908796f08ae8e00d4c8fc330a3f/cidc_schemas/prism/merger.py#L139

That said, it's probably better not to prematurely optimize here.

jacoblurye (approved):
With some direct tests for `_remove_optional_uuid_recursive`, I think this should be good to go.

cidc_api/resources/upload_jobs.py
```diff
@@ -165,6 +177,40 @@ def update_upload_job(upload_job: UploadJobs, upload_job_updates: UploadJobs):
...
 
 
+def _remove_optional_uuid_recursive(target: dict, uuid: str):
+    """
+    If target contains an item : dict with {"upload_placeholder":uuid}, removes that item and returns the modified target
+    If no such item is found, continues recursively as depth first search
+    If the uuid is never found, returns the target unchanged
+    """
+    if isinstance(target, dict):
+        if "upload_placeholder" in target and target["upload_placeholder"] == uuid:
```
jacoblurye, line 187: This can be shortened to `target.get("upload_placeholder") == uuid`.

scvannost, line 187:
```suggestion
        if target.get("upload_placeholder") == uuid:
```

scvannost: in next commit

cidc_api/resources/upload_jobs.py
scvannost: yeah, this isn't optimized at all. but hopefully it won't be an issue, otherwise I think we should leave it as "tech debt" for now

PR #374, opened 2021-01-19
Add plotly dash proof-of-concept

jacoblurye:
This PR adds a one-off example of an auth-protected Plotly Dash dashboard to the API. In order to authenticate Dash requests, I had to update our `auth` module to accept user identity tokens via the JSON body of client requests in addition to the `Authorization` header.

If we decide to move forward with Plotly Dash, I intend to make some kind of microframework (or just a set of helper methods) to simplify the process of adding new dashboards to the API in a future PR.

See corresponding UI PR: https://github.com/CIMAC-CIDC/cidc-ui/pull/345

### DIFF
--- a/cidc_api/dash_example.py
+++ b/cidc_api/dash_example.py
@@ -1,0 +1,125 @@
+import dash
+import dash_table as dt
+import dash_core_components as dcc
+import dash_html_components as html
+from dash.dependencies import Output, Input
+from flask import Flask, request
+from sqlalchemy.sql.selectable import Select
+
+from cidc_api.models import (
+    UploadJobSchema,
+    UploadJobStatus,
+    UploadJobs,
+    CIDCRole,
+    TrialMetadata,
+)
+from cidc_api.config.settings import MAX_PAGINATION_PAGE_SIZE
+from cidc_api.shared.auth import requires_auth
+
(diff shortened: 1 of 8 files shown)


cidc_api/dash_example.py
```diff
@@ -0,0 +1,125 @@
...
+        index_string=index_string,
+    )
+    dash_app.layout = layout
+
+    @dash_app.callback(Output(TRIAL_ID_DROPDOWN_ID, "options"), NoInputs)
+    @requires_auth("dash.trial_ids", [CIDCRole.ADMIN.value])
+    def populate_trial_ids(_):
+        """Load available trial_ids into the trial id dropdown."""
+        return [
+            {"label": t, "value": t} for t in TrialMetadata.get_distinct("trial_id")
```
curlup, line 100: nit: do you really need `distinct` here? I'd think it would be just list of trial ids, which are unique by the db constraint

curlup (approved):
lgtm

cidc_api/dash_example.py
jacoblurye: True that `distinct` isn't necessary, but it seemed worth it to use our predefined helper method for the sake of code brevity.

PR #381, opened 2021-01-26
Create an intake bucket per user

jacoblurye:
...instead of granting uploaders conditional access on subdirectories of a single intake bucket. 

This change will allow users to upload their files directly via the GCS console UI and should fix IAM issues with the current implementation.

This PR contains breaking changes to the `/ingestion/intake_gcs_uri` endpoint and should be merged alongside https://github.com/CIMAC-CIDC/cidc-ui/pull/348.

### DIFF
--- a/cidc_api/shared/gcloud_client.py
+++ b/cidc_api/shared/gcloud_client.py
@@ -3,6 +3,7 @@ import json
 import datetime
 import re
 import warnings
+import hashlib
 from collections import namedtuple
 from concurrent.futures import Future
 from typing import List, Tuple, Optional, BinaryIO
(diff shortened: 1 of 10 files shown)


cidc_api/shared/gcloud_client.py
```diff
@@ -149,92 +134,73 @@ def revoke_upload_access(user_email: str):
...
-    prefix = _build_intake_prefix(user_id, user_email, trial_id, upload_type)
-
-    logger.info(f"Granting intake access on {prefix} to {user_email}")
-
-    return grant_conditional_gcs_access(
-        GOOGLE_INTAKE_BUCKET, prefix, GOOGLE_INTAKE_ROLE, user_email
-    )
+    # 10 characters should be plenty, given that we only expect
+    # a handful of unique data uploaders - we get 16^10 possible hashes.
+    email_hash = hashlib.sha1(bytes(user_email, "utf-8")).hexdigest()[:10]
```
curlup, line 144: nit: no salt though?

curlup (approved):

cidc_api/shared/gcloud_client.py
jacoblurye: Per our offline discussions, I'm not going to add a salt, since the main goal of hashing emails is to avoid directly publishing user information in bucket URLs, not to obscure which bucket belongs to which user.

PR #383, opened 2021-02-02
Re-organize Dash example and add dashboard tests

jacoblurye:
This PR more formally integrates the Dash proof-of-concept from #374 into the API code and provides an example of testing a Dash dashboard.

New CIDC dashboards should be created using the `create_new_dashboard` utility then attached to the API Flask app in the `register_dashboards` function.


cidc_api/shared/auth.py
```diff
@@ -322,27 +322,17 @@ def _enforce_cli_version():
         logger.error(f"Unrecognized user-agent string format: {user_agent}")
         raise BadRequest("could not parse User-Agent string")
 
-    # Old CLI versions don't update the User-Agent header, so we (perhaps dangerously)
-    # assume any request coming from the python requests library is from a "very" old
-    # version of the CLI.
-    is_very_old_cli = client == "python-requests"
```
jacoblurye, line 328: The dash testing utility uses the python requests library, so this check was causing problems. Since this check only applied to very, very old versions of the CLI that didn't set the user-agent header, I think it's fine to remove it.

scvannost (approved):

PR #395, opened 2021-02-23
Convert upload jobs table into a shipments dashboard

jacoblurye:
Replace the `upload_jobs_table` admin dashboard with a dashboard for inspecting sample shipments that have been successfully ingested into the CIDC. Adapted from @jim-bo's ipython notebook prototype.


jim-bo (approved):
So if I understand this: I think you've answered the de-duplication problem by sorting the upload jobs by when the merge was completed, then grabbing the most recent.

I agree with that strategy and you've nicely leveraged SQLalchemy instead of pandas. Seems like a good template.

PR #409, opened 2021-03-11
Add new 'network-viewer' role to RBAC

jacoblurye:
`network-viewer`s can list metadata / what files are available for trials and assays that they have permission to view. However, they are not given permission to download any files from GCS.

### DIFF
--- a/cidc_api/models/models.py
+++ b/cidc_api/models/models.py
@@ -415,8 +415,12 @@ class Permissions(CommonColumns):
                 orig=f"`granted_by_user` user must exist, but no user found with id {self.granted_by_user}",
             )
 
+        is_network_viewer = grantee.role == CIDCRole.NETWORK_VIEWER.value
+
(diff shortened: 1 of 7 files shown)


cidc_api/models/models.py
```diff
@@ -432,6 +437,10 @@ def insert(self, session: Session, commit: bool = True, compute_etag: bool = Tru
         # Always commit, because we don't want to grant IAM download unless this insert succeeds.
         super().insert(session=session, commit=True, compute_etag=compute_etag)
```
scvannost, line 438: does this want to be after the return of network viewers?

scvannost (approved):

cidc_api/models/models.py
jacoblurye: No - this code inserts the permission record into the database, which is what gives a network viewer permission to view metadata for files pertaining to a certain trial/upload type (but not to download them).

PR #416, opened 2021-03-17
Add facets for nanostring

scvannost:


jacoblurye (approved):
If you want to add descriptions for these facets as a second argument to `FacetConfig`, that would be nice to have - not crucial, though.

PR #422, opened 2021-03-23
Add data access report generation to the Users model

jacoblurye:
This PR adds functionality to create an automatic spreadsheet report on what users have access to what data types on what trials, leaning on pandas [`ExcelWriter`](https://pandas.pydata.org/docs/reference/api/pandas.ExcelWriter.html) feature.

Currently working on a UI PR to add a button for triggering/downloading these reports from the portal admin page.

### DIFF
--- a/cidc_api/models/models.py
+++ b/cidc_api/models/models.py
@@ -1,10 +1,13 @@
 import re
+import csv
 import hashlib
 from datetime import datetime, timedelta
 from enum import Enum as EnumBaseClass
 from functools import wraps
-from typing import Dict, Optional, List, Union, Callable, Tuple
+from io import BytesIO
+from typing import BinaryIO, Dict, Optional, List, Union, Callable, Tuple
 
+import pandas as pd
 from flask import current_app as app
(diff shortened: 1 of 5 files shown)


cidc_api/models/models.py
```diff
@@ -345,6 +349,54 @@ def disable_inactive_users(session: Session, commit: bool = True):
...
+        Save an excel file to the given file handler, and return the pandas dataframe
+        used to generate that excel file.
+        """
+        user_columns = (Users.email, Users.organization, Users.role)
+
+        query = (
+            session.query(
+                *user_columns,
+                Permissions.trial_id,
+                func.string_agg(Permissions.upload_type, ","),
```
curlup, line 370: nit: i'd go with `array_agg` and do formatting in python instead

curlup (approved):

cidc_api/models/models.py
jacoblurye: Ah, sure - I think I'm going to forgo this for now until there's indication we need the permissions info in a more flexible format.

PR #431, opened 2021-03-31
Add method and API endpoint for summarizing data ingested for each trial

jacoblurye:
This PR adds the `/trial_metadata/summaries` endpoint, which returns total file size and assay-sample counts for each trial in the database.

### DIFF
--- a/cidc_api/config/settings.py
+++ b/cidc_api/config/settings.py
@@ -27,8 +27,6 @@ assert ENV == "dev" if DEBUG else True, "DEBUG mode is only allowed when ENV='de
 TESTING = environ.get("TESTING") == "True"
 ALLOWED_CLIENT_URL = environ.get("ALLOWED_CLIENT_URL")
 IS_GUNICORN = "gunicorn" in environ.get("SERVER_SOFTWARE", "")
-# DO NOT CHANGE - used only for mocking in tests that need to bypass this feature
-DISABLE_METADATA_VALIDATION_ON_INSERT = False
 
 ### Configure miscellaneous constants ###
 TEMPLATES_DIR = path.join("/tmp", "templates")
--- a/cidc_api/config/settings.py
+++ b/cidc_api/config/settings.py
@@ -27,6 +27,8 @@ assert ENV == "dev" if DEBUG else True, "DEBUG mode is only allowed when ENV='de
 TESTING = environ.get("TESTING") == "True"
 ALLOWED_CLIENT_URL = environ.get("ALLOWED_CLIENT_URL")
 IS_GUNICORN = "gunicorn" in environ.get("SERVER_SOFTWARE", "")
+# DO NOT CHANGE - used only for mocking in tests that need to bypass this feature
+DISABLE_METADATA_VALIDATION_ON_INSERT = False
 
 ### Configure miscellaneous constants ###
 TEMPLATES_DIR = path.join("/tmp", "templates")
--- a/tests/models/test_models.py
+++ b/tests/models/test_models.py
@@ -367,6 +367,84 @@ def test_partial_patch_trial_metadata(clean_db):
 
 
 @db_test
+def test_trial_metadata_get_summaries(clean_db, monkeypatch):
+    """Check that trial data summaries are computed as expected"""
+    monkeypatch.setattr(
+        "cidc_api.models.models.DISABLE_METADATA_VALIDATION_ON_INSERT", True
+    )
+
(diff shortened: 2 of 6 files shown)


tests/models/test_models.py
```diff
@@ -1101,4 +1179,4 @@ def test_user_get_data_access_report(clean_db, monkeypatch):
         if user == admin_user:
             assert set(["*"]) == set(user_df.permissions)
         else:
-            assert set(["wes,cytof"]) == set(user_df.permissions)
+            assert set(user_df.permissions) - set(["wes,cytof", "cytof,wes"]) == set()
```
jacoblurye, line 1182: This is an unrelated fix for a test that I found fails non-deterministically.

cidc_api/config/settings.py
```diff
@@ -27,6 +27,8 @@
 TESTING = environ.get("TESTING") == "True"
 ALLOWED_CLIENT_URL = environ.get("ALLOWED_CLIENT_URL")
 IS_GUNICORN = "gunicorn" in environ.get("SERVER_SOFTWARE", "")
+# DO NOT CHANGE - used only for mocking in tests that need to bypass this feature
+DISABLE_METADATA_VALIDATION_ON_INSERT = False
```
scvannost, line 31: This doesn't seem like the best way to me; seems kind of hacky to have functionality that's only needed for testing

tests/models/test_models.py
```diff
@@ -366,6 +366,84 @@ def test_partial_patch_trial_metadata(clean_db):
...
+    """Check that trial data summaries are computed as expected"""
+    monkeypatch.setattr(
+        "cidc_api.models.models.DISABLE_METADATA_VALIDATION_ON_INSERT", True
+    )
+
+    # Add some trials
+    records = [{"fake": "record"}]
+    tm1 = {
+        **METADATA,
+        "protocol_identifier": "tm1",
```
scvannost, line 380: maybe add a comment saying that this is purposefully overriding the `protocol_identifier` in `METADATA`

tests/models/test_models.py
scvannost: there's also https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertSetEqual

tests/models/test_models.py
```diff
@@ -366,6 +366,84 @@ def test_partial_patch_trial_metadata(clean_db):
     TrialMetadata._patch_trial_metadata(TRIAL_ID, metadata_patch)
 
 
+@db_test
+def test_trial_metadata_get_summaries(clean_db, monkeypatch):
+    """Check that trial data summaries are computed as expected"""
+    monkeypatch.setattr(
+        "cidc_api.models.models.DISABLE_METADATA_VALIDATION_ON_INSERT", True
+    )
```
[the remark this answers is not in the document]
jacoblurye: Based on what I tried before taking the hacky approach, no - maybe it was user error on my part, but it may also be because of how sqlalchemy's `@validates` decorator works. It's possible that by looking into sqlalchemy implementation internals I'll be able to find what needs to be mocked, though.

tests/models/test_models.py
jacoblurye: Good to know! We never expect set equality here, though, since `users_df` only has one row.

scvannost (approved):

cidc_api/config/settings.py
curlup: agreed
@jacoblurye you can monkey patch the code while in tests to circumvent the validation.

tests/models/test_models.py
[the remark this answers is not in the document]
curlup: one way to disentangle from `@validates` is to introduce one extra indirection layer, something like:
```
@validates
def validate_metadata_json_on_insert(...):
    return _acutally_validate_mdjson(...)

def _acutally_validate_mdjson(...)
```
and then monkeypatch `_acutally_validate_mdjson`

tests/models/test_models.py
curlup: nit pick but something like
```suggestion
            assert set(user_df.permissions) < set(["wes,cytof", "cytof,wes"])
```

curlup (approved):

tests/models/test_models.py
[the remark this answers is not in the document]
jacoblurye: Oh, of course! Should've thought of that 😅 Thanks @curlup.

PR #433, opened 2021-04-01
Bump for schemas

scvannost:
see https://github.com/CIMAC-CIDC/cidc-schemas/pull/437

### DIFF
--- a/cidc_api/models/files/details.py
+++ b/cidc_api/models/files/details.py
@@ -435,8 +435,8 @@ details_dict = {
     ),
     "npx|analysis_ready|csv": FileDetails(
         "analysis",
-        "comma-separated table of Normalized Protein eXpressions for the entire study",
-        "The comma-separated, plain-text table of Normalized Protein eXpressions for all samples and analytes across the entire study.\nEach row is a sample by CIMAC ID and each column is a sample by (Name, Uniprot ID, *Olink ID*).",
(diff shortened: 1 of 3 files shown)


cidc_api/models/files/details.py
```diff
@@ -435,8 +435,8 @@ class FileDetails(NamedTuple):
     ),
     "npx|analysis_ready|csv": FileDetails(
         "analysis",
-        "comma-separated table of Normalized Protein eXpressions for the entire study",
-        "The comma-separated, plain-text table of Normalized Protein eXpressions for all samples and analytes across the entire study.\nEach row is a sample by CIMAC ID and each column is a sample by (Name, Uniprot ID, *Olink ID*).",
+        "comma-separated table of Normalized Protein eXpressions for for all analytes/samples across the entire study",
```
jim-bo, line 438: Hey, we can leave it out for now, or put a ticket in the backlog but I think we should put a bit more meat into the descriptions of these analysis ready files. Describing that empty values represent a failed assay, you can use LOD to further filter values which are questionable. There is probably a lot we could do here so now that I've finished typing I would suggest we put a ticket in the backlog.

jim-bo (approved):

cidc_api/models/files/details.py
scvannost: in the backlog!

PR #436, opened 2021-04-01
Fix API for new CyTOF DM; And file details/facet configs for ELISA

scvannost:


cidc_api/models/files/facets.py
```diff
@@ -236,6 +236,7 @@ def __init__(self, match_clauses: List[str], description: Optional[str] = None):
             "Data files from TCRseq analysis indicating TRA & TRB clones UMI counts",
         ),
     },
+    "ELISA": {"All Files": FacetConfig(["/elisa/assay.xlsx"])},
```
jacoblurye, line 239: Would it be appropriate to label this facet `"Source"`? Or, if we only expect one facet for `"ELISA"` going forward, we could use a non-nested facet like this:
```python
"ELISA": FacetConfig(["/elisa/assay.xlsx"])
```
(see the `clinical_facets` as examples below)

jacoblurye (approved):

cidc_api/models/files/facets.py
scvannost: would [olink](https://github.com/CIMAC-CIDC/cidc-api-gae/blob/master/cidc_api/models/files/facets.py#L206) also want this type of labeling?

jacoblurye: Maybe - we might want multiple nested facets to distinguish between the "raw" npx and ct files and the combined biomarker file you're introducing.

scvannost: in next commit

scvannost:
will also close https://github.com/CIMAC-CIDC/cidc-api-gae/pull/435

cidc_api/models/files/details.py
```diff
@@ -12,62 +12,118 @@ class FileDetails(NamedTuple):
...
         "A plain-text, comma-separated table with a column for each profiled subset of all assigned cell types, where rows are the signal on each channel for every profiled subset.",
     ),
-    "/cytof_analysis/source.fcs": FileDetails(
+    "/cytof_10021_analysis/source.fcs": FileDetails(
         "source",
         "fcs data used as the input for this analysis",
         "The analysis-ready FCS file used as the input for this analysis. After normalization, debarcoding, and removal of Veri-Cells and other non-specimen cells.",
     ),  # schemas/assays/components/cytof/cytof_analysis.json#properties/fcs_file
+    # CyTOF E4412
+    "/cytof_e4412/spike_in.fcs": FileDetails(
```
jacoblurye, line 72: I wonder if it's worth factoring out identical `FileDetails` definitions across cytof variants into shared variables (so we that we don't need to edit descriptions in two places, e.g.).

cidc_api/models/files/facets.py
```diff
@@ -204,17 +219,18 @@ def __init__(self, match_clauses: List[str], description: Optional[str] = None):
         ),
     },
     "Olink": {
-        "All Olink Files": FacetConfig(
+        "Run Level": FacetConfig(
```
jacoblurye, line 222: Maybe hyphenate? `Run-Level` and `Study-Level`.

cidc_api/models/files/facets.py
```diff
@@ -247,9 +264,12 @@ def __init__(self, match_clauses: List[str], description: Optional[str] = None):
     ),
 }
 
+analysis_ready_facets = {"Olink": FacetConfig(["npx|analysis_ready|csv"])}
```
jacoblurye, line 267: I'm not sure about adding this entirely separate facet category - maybe we should just leave this file available in the `"Olink" > "Study-Level"` facet, with a plan to discuss as a team how we want to better surface and describe CIDC-generated derivative files in the portal?

cidc_api/models/models.py
```diff
@@ -945,6 +945,38 @@ def get_summaries(session: Session) -> List[dict]:
...
+                trial_id,
+                'cytof' as key,
+                sum(jsonb_array_length(batches->'records')) as value
+            from
+                trial_metadata,
+                jsonb_each(metadata_json->'assays') assays,
+                jsonb_array_elements(value) batches
+            where key in ('cytof_10021', 'cytof_e4412')
+            group by trial_id
+        """
```
jacoblurye, line 978: Since we're just renaming the `cytof_10021` and `cytof_e4412` assay types to `cytof` here, I'm not convinced this needs a separate subquery - I think we should only create separate subqueries for assays that deviate from the typical "batches of sample-level records" shape. 

Instead, I'd prefer if we introduce this change (and other assay renamings, like `'hande' -> 'h&e'`) in the `generic_assay_subquery` like:
```python
generic_assay_subquery = """
    select
        trial_id,
        case
            when key in ('cytof_e4412', 'cytof_10021') then 'cytof'
            when key = 'hande' then 'h&e'
            else key
        end as key,
        sum(jsonb_array_length(batches->'records')) as value
    from
        trial_metadata,
        jsonb_each(metadata_json->'assays') assays,
        jsonb_array_elements(value) batches
    where key not in ('olink', 'nanostring')
    group by trial_id, key
"""
```

scvannost: oops, I forgot to add the different handling for e4412 because there's a participants level to go down too

jacoblurye: Ooo, hang on. I don't think the `group by key` will work in the example code I wrote above (pretty sure the different `cytof_*`s will be grouped separately). I do still wonder if there's a no-extra-subquery way to do it, though.

cidc_api/models/files/facets.py
scvannost: it's also there, this was just a suggestion that I had as a good place to put this saliently for the users

cidc_api/models/models.py
scvannost: I can do this for hande and cytof_10021, then just have a subquery for cytof_e4412 because of the different structure?

cidc_api/models/files/facets.py
scvannost: in next commit

cidc_api/models/files/details.py
scvannost: yeah, that's not too bad. in next commit

cidc_api/models/files/facets.py
jacoblurye: Yeah, on second thought, let's just go for this. Better to put it somewhere people will stumble across it and make changes if necessary based on their feedback.

cidc_api/models/models.py
jacoblurye: [Resolved offline - subsequent solution in the PR represents what we discussed]

PR #439, opened 2021-04-07
Update TrialMetadata.list to prune raw metadata

jacoblurye:
We've recently noted that the endpoint for listing trials, `GET /trial_metadata`, has been running very slowly. Local profiling suggests that we can speed up this trial listing operation by pruning unneeded branches of the raw trial metadata JSON. 

So, this PR updates `TrialMetadata.list` to issue a database query that does not load the `participants`, `assays`, `analysis`, and `shipments` branches of each trial's metadata blob, saving substantially on the amount of data this query loads.

Additionally, this PR adds a URL query param `include_counts` to the `GET /trial_metadata` endpoint which, when true, indicates to the API that it should include trial-level participant and sample counts for each trial (since the raw metadata JSON will no longer include the info required for API clients to compute this themselves). 

This should be merged *after* a corresponding UI PR that handles this new endpoint behavior.

### DIFF
--- a/cidc_api/models/models.py
+++ b/cidc_api/models/models.py
@@ -793,27 +793,124 @@ class TrialMetadata(CommonColumns):
         return unprism.unprism_samples(trial.metadata_json)
 
     file_bundle: Optional[FileBundle]
+    num_participants: Optional[int]
+    num_samples: Optional[int]
+
+    @classmethod
+    def _pruned_metadata_json(cls):
+        """
+        Builds a modified metadata_json column selector with the "assays", "analysis",
+        "shipments", and "participants" properties removed.
+        """
+        return (
+            func.jsonb_set(
+                cls.metadata_json, "{participants}", literal_column("'[]'::jsonb")
+            )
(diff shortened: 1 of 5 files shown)


jim-bo:
Not sure I can provide great feedback on the database queries, but I think the structure is sound. Do you have any evidence that this improves the performance though? I don't think we need tests around it, just an assertion should be fine.

jim-bo (approved):
Seems good, but does it improve performance?

jacoblurye:
Sure, @jim-bo! Apologies - I should've included the performance data in my PR description. With a local API instance pointed at the production database, here's the relative performance between current and updated implementations (N=10):

current: `GET /trial_metadata?include_file_bundles=true` sends `30MB` of data in `9.62 ± 1.06`.

updated: `GET /trial_metadata?include_file_bundles=true&include_counts=true` sends `593KB` of data in `2.60 ± 0.33`.

scvannost (approved):

PR #444, opened 2021-04-12
Companion prod deploy to keep version synced with CFns

scvannost:
[0.24.14 to 0.24.15](https://github.com/CIMAC-CIDC/cidc-api-gae/pull/442/files)
https://github.com/CIMAC-CIDC/cidc-cloud-functions/pull/214


scvannost:
don't need to do this, and further updates to master are more important than keeping in line

PR #446, opened 2021-04-12
Make data_format and file_name columns optional

jacoblurye:
...in preparation for removing these fields from the JSON schema data model. Also, remove from the test data, where possible.

We should drop these columns from the database entirely at some point, but the only strong reason for doing that is preventing future developer confusion.


roshni-b (approved):
👍🏼

PR #449, opened 2021-04-14
Add clinical data participant counts to the trial summary query

jacoblurye:


jim-bo (approved):
I guess no real comment, it seems too simple?

scvannost (approved):

PR #454, opened 2021-04-20
Add support for trial- and assay-level permissions

jacoblurye:
This PR makes it possible for admins to grant any user access to all data across assays for a trial or to all data across trials for an assay.

### DIFF
--- a/cidc_api/models/models.py
+++ b/cidc_api/models/models.py
@@ -169,7 +169,6 @@ class CommonColumns(BaseModel):  # type: ignore
         """List records in this table, with pagination support."""
         query = session.query(cls)
         query = cls._add_pagination_filters(query, **pagination_args)
-        print(query)
         return query.all()
 
     @classmethod
--- a/cidc_api/models/models.py
+++ b/cidc_api/models/models.py
@@ -551,7 +551,7 @@ class Permissions(CommonColumns):
         # If a delete operation fails, all other deletes and the insertion will
         # be rolled back.
         for perm in perms_to_delete:
-            perm.delete(deleted_by=self.granted_by_user, commit=False)
+            session.delete(perm)
 
         # Always commit, because we don't want to grant IAM download unless this insert succeeds.
         super().insert(session=session, commit=True, compute_etag=compute_etag)
--- a/cidc_api/models/models.py
+++ b/cidc_api/models/models.py
@@ -33,6 +33,8 @@ from sqlalchemy import (
     literal_column,
     not_,
     literal,
+    and_,
+    or_,
 )
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.hybrid import hybrid_property
@@ -407,6 +409,17 @@ class IAMException(Exception):
     pass
 
 
+EXTRA_DATA_TYPES = ["participants info", "samples info"]
+ALL_UPLOAD_TYPES = set(
+    [
+        *prism.SUPPORTED_MANIFESTS,
+        *prism.SUPPORTED_ASSAYS,
+        *prism.SUPPORTED_ANALYSES,
+        *EXTRA_DATA_TYPES,
+    ]
+)
+
+
 class Permissions(CommonColumns):
     __tablename__ = "permissions"
     __table_args__ = (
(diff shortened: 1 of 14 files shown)


cidc_api/models/models.py
```diff
@@ -166,6 +169,7 @@ def list(cls, session: Session, **pagination_args):
         """List records in this table, with pagination support."""
         query = session.query(cls)
         query = cls._add_pagination_filters(query, **pagination_args)
+        print(query)
```
curlup, line 172: hm

cidc_api/models/models.py
```diff
@@ -488,9 +523,37 @@ def insert(self, session: Session, commit: bool = True, compute_etag: bool = Tru
...
+                else Permissions.upload_type == self.upload_type,
+            )
+            .all()
         )
 
+        # Add any related permission deletions to the insertion transaction.
+        # If a delete operation fails, all other deletes and the insertion will
+        # be rolled back.
+        for perm in perms_to_delete:
+            perm.delete(deleted_by=self.granted_by_user, commit=False)
```
curlup, line 555:
```suggestion
            perm.delete(deleted_by=self.granted_by_user, session=session, commit=False)
```

cidc_api/models/models.py
```diff
@@ -501,7 +564,13 @@ def insert(self, session: Session, commit: bool = True, compute_etag: bool = Tru
         try:
             # Grant IAM permission in GCS only if db insert worked
             grant_download_access(grantee.email, self.trial_id, self.upload_type)
+            # Remove permissions staged for deletion, if any
+            for perm in perms_to_delete:
+                revoke_download_access(grantee.email, perm.trial_id, perm.upload_type)
         except Exception as e:
+            # Add back deleted permissions, if any
+            for perm in perms_to_delete:
+                perm.insert()
```
curlup, line 573: again maybe `session=session`?

cidc_api/models/models.py
```diff
@@ -501,7 +564,13 @@ def insert(self, session: Session, commit: bool = True, compute_etag: bool = Tru
...
             # Delete the just-created permissions record
             super().delete()
```
curlup, line 575: don't you want to just roll back current session / transaction?

cidc_api/models/models.py
```diff
@@ -488,9 +523,37 @@ def insert(self, session: Session, commit: bool = True, compute_etag: bool = Tru
...
         logger.info(
-            f"admin-action: {grantor.email} gave {grantee.email} the permission {self.upload_type} on {self.trial_id}"
+            f"admin-action: {grantor.email} gave {grantee.email} the permission {self.upload_type or 'all assays'} on {self.trial_id or 'all trials'}"
+        )
+
+        # If this is a permission granting the user access to all trials for
+        # a given upload type or all upload types for a given trial, delete
+        # any related trial-upload type specific permissions to avoid
+        # redundancy in the database and in conditional IAM bindings.
+        perms_to_delete = (
```
curlup, line 533: Nit: i see the point of cleaning up IAM bindings, not absolutely sure about the db cleanup.

migrations/versions/a0f25824b2ae_permissions_nullable_trial_and_upload_.py
```diff
@@ -0,0 +1,40 @@
...
+        "ck_nonnull_trial_id_or_upload_type",
+        "permissions",
+        "trial_id is not null or upload_type is not null",
+    )
+
+
+def downgrade():
+    op.drop_constraint("ck_nonnull_trial_id_or_upload_type", "permissions")
+    op.alter_column(
+        "permissions", "upload_type", existing_type=sa.VARCHAR(), nullable=False
```
curlup, line 36: potentially needs a corresponding `delete` first - or rollback might fail due to existing records with `field = null`

curlup (approved):

cidc_api/models/models.py
jacoblurye: I think it gets harder to reason about correctness when we don't enforce a 1-1 correspondence between database permission records and IAM permission.

cidc_api/models/models.py
jacoblurye: Since we loaded `perms_to_delete` via `session.query`, I think each instance is already bound to `session`. I'll double check this, though.

cidc_api/models/models.py
jacoblurye: The transaction has already been committed here https://github.com/CIMAC-CIDC/cidc-api-gae/pull/454/files#diff-695d07bda541617439f1fc301a4ea5b1794d003d385e0fd1361d920e060f7ab6R558, so I don't think it can be rolled back.

Some context for this: we commit the permission record insert to ensure we catch foreign key / uniqueness / etc. issues before creating a corresponding IAM permission.

cidc_api/models/models.py
curlup: well for the readability sake i'd still add it

cidc_api/models/models.py
curlup: right, then what do you think on not implicitly deleting those (granted by someone for some reason) explicit permissions at all? i understand there's a limit on the IAM bindings though. anyways it's a minor thing.

cidc_api/models/models.py
jacoblurye: took another, similar route (`session.delete(perm)`) - the readability argument is a good one!

PR #461, opened 2021-04-26
Update facet_group for lane support

scvannost:


cidc_api/models/files/details.py
```diff
@@ -23,6 +23,16 @@ class FileDetails(NamedTuple):
...
     ),
+    "/wes/r1_L.fastq.gz": FileDetails(
+        "source",
+        "fastq file of raw Read 1, compressed",
+        "The gzipped, FASTQ file that represents the 5' read of paired sequencing. Generated by the sequencing machine defined by the RNA assay this file is associated with.",
+    ),
+    "/wes/r2_L.fastq.gz": FileDetails(
+        "source",
+        "fastq file of raw Read 2, compressed",
+        "The gzipped, FASTQ file that represents the 3' read of paired sequencing. Generated by the sequencing machine defined by the RNA assay this file is associated with.",
```
jacoblurye, line 34: Should this be `5'`, as with `r1_L`?

jacoblurye (approved):

cidc_api/models/files/details.py
scvannost: nope, read 1 is the 5' end while read 2 is the 3' end

PR #466, opened 2021-04-28
Update file details and facets for TCRseq

roshni-b:
- Update TCR analysis facets so they start to show up in the portal
- Add tcr_analysis/report_trial.tar.gz to file list (ref: https://github.com/CIMAC-CIDC/cidc-schemas/pull/399)
- API version bump


scvannost (approved):
you don't need to bump the version because they're not used in cidc-api-modules

PR #470, opened 2021-05-03
Update gunicorn and app engine configs to improve performance

jacoblurye:
This PR does two related things:
* Updates `gunicorn.conf.py` config to use the coroutine-based [`gevent` worker type](https://docs.gunicorn.org/en/stable/design.html#async-workers). A single `gevent` worker can handle many concurrent requests at once.
* Updates `app.*.yaml` to permit more concurrent requests and higher CPU load per app engine instance before autoscaling up additional instances.

These config changes allow us to get more performance with the same resources, leading to a noticeable performance improvement on Portal pages that send many concurrent requests to the API. For example, whereas loading the entire `/profile` page under the current API config takes `10.08s` on average, under the updated configuration it takes `4.10s` (N=5, running the staging portal on Chrome).

### DIFF
--- a/gunicorn.conf.py
+++ b/gunicorn.conf.py
@@ -1,6 +1,5 @@
 """Configuration for the gunicorn WSGI server."""
 import os
-import multiprocessing
 
 from gevent.monkey import patch_all
 from psycogreen.gevent import patch_psycopg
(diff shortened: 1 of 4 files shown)


scvannost (approved):

roshni-b (approved):

gunicorn.conf.py
```diff
@@ -1,13 +1,25 @@
...
+# Use async workers: https://docs.gunicorn.org/en/stable/design.html#async-workers
+worker_class = "gevent"
 # See https://docs.gunicorn.org/en/stable/settings.html
 port = os.environ.get("PORT", 8080)
-loglevel = "DEBUG" if is_dev else "INFO"
-reload = is_dev
+# Development mode - restart the server on code changes
+reload = os.environ.get("ENV") == "dev"
+# Include debug level-logs if we're in development mode
+loglevel = "DEBUG" if reload else "INFO"
```
curlup, line 21: unrelated, but: not sure why not have DEBUG for all envs.

curlup (approved):

gunicorn.conf.py
jacoblurye: I think this is because gunicorn debug-level logs dump a lot of info that generally hasn't proven helpful. I can't remember an instance where we've encountered a bug on staging or prod whose root cause was masked by excluding these logs.

PR #490, opened 2021-05-20
prod deploy: updated factes, fixed facet counts, WES tumor-only support, CyTOF s1609/gd2car support

roshni-b:


scvannost (approved):
should also get review from @jacoblurye

jacoblurye (approved):

PR #495, opened 2021-05-25
Data overview: add analysis sample counts and excluded sample info to TrialMetadata.get_summary

jacoblurye:


cidc_api/models/models.py
```diff
@@ -1265,6 +1265,136 @@ def get_summaries(session: Session) -> List[dict]:
...
+                trial_metadata,
+                jsonb_array_elements(metadata_json#>'{assays,cytof_10021}') batch,
+                jsonb_array_elements(batch->'records') record
+        """
+
+        wes_analysis_subquery = """
+            select
+                trial_id,
+                'wes_analysis' as key,
+                2 * jsonb_array_length(metadata_json#>'{analysis,wes_analysis,pair_runs}') as value
```
scvannost, line 1299: this will also count anything with only a tumor/normal pairing, as that manifest generates minimal wes_analysis pair_runs
seems like we would want to check if the other subfields are not null

roshni-b (approved):

PR #512, opened 2021-06-28
CyTOF s1609/gd2car specific subquery

roshni-b:


cidc_api/models/models.py
```diff
@@ -1214,7 +1214,7 @@ def get_summaries(session: Session) -> List[dict]:
             select
                 trial_id,
                 case
-                    when key = 'cytof_10021_9204' then 'cytof'
+                    when key in ('cytof_10021_9204', 'cytof_s1609_gd2car') then 'cytof'
```
scvannost, line 1217: I think you only need one of this or the specific query

PR #549, opened 2021-08-11
Step 1 of Relational DB towards CSMS Integration

scvannost:
Initial set up of tables and definition of needed classes for base metadata and assay uploads
    Manifests: pbmc, tissue_slide; demo for clinical_data
    Assays: h_and_e, wes_[fastq/bam]
Generated new-style templates and added full testing data
Implemented JSON -> Relational sync function and wired for testing

TO DO as of EOD 11Aug:
* way to trigger initial synchronization, as this is needed to test maintaining alignment


cidc_api/models/templates/sync_schemas.py
```diff
@@ -0,0 +1,153 @@
...
+
+        for partic in upload.metadata_patch["participants"]:
+            for sample in partic["samples"]:
+                sample_to_shipment_map[sample["cimac_id"]] = manifest_id
+
+    return sample_to_shipment_map
+
+
+@with_default_session
+def syncall_from_blobs(session: Session, dry_run: bool = False,) -> List[Exception]:
```
jim-bo, line 39: @scvannost so this function seem wayyy long. It'll be hard to provide meaningful feedback on this without inline comments. Ideally this would be broken up into multiple functions with doc-strings describing the purpose.

cidc_api/models/templates/utils.py
```diff
@@ -0,0 +1,164 @@
...
+    return set(cls.__bases__).union([s for c in cls.__bases__ for s in _all_bases(c)])
+
+
+def _all_subclasses(cls: Type) -> Set[Type]:
+    return set(cls.__subclasses__()).union(
+        [s for c in cls.__subclasses__() for s in _all_subclasses(c)]
+    )
+
+
+def _get_global_insertion_order() -> List[MetadataModel]:
```
jim-bo, line 20: Gonna defer this to @curlup or perhaps this was already reviewed as part of a previous commit?

jim-bo (changes requested):
Sorry for the short review but I think more comments, and breaking up the sync function is necessary before I can grok it. It would also be good to highlight somewhere in the PR if some of the other changes (such as around figuring out insertion order of DB queries) have already been reviewed.

cidc_api/models/templates/sync_schemas.py
scvannost: in next commit

scvannost:
In the backlog are tickets to
* Add GCS file storage handling
* Handle storing the template itself
* Create any developer documentation

cidc_api/models/templates/sync_schemas.py
```diff
@@ -0,0 +1,226 @@
...
+    Uses the UploadJobs where each patch contains only a single shipment and its samples.
+    """
+    sample_to_shipment_map = {}
+
+    uploads = (
+        session.query(UploadJobs)
+        .filter(UploadJobs.trial_id == trial_id, UploadJobs.multifile == False)
+        .all()
+    )
+    for upload in uploads:
```
jim-bo, line 28: @scvannost How does this deal with multiple uploads of the same shipping manifest which we see frequently in our existing database? Just curious.

cidc_api/models/templates/sync_schemas.py
```diff
@@ -0,0 +1,226 @@
...
+            errors.append(e)
+
+    return errors
+
+
+def _sync_shipments(
+    trial: TrialMetadata, new_trial: ClinicalTrial, session: Session
+) -> List[Exception]:
+    """
+    Given corresponding TrialMetadata and ClinicalTrial instances, copy over shipments.
```
jim-bo, line 150: Similar to a previous comment: How do you deal with the trials in-which there are multiple shipping manifest uploads (even if if they are the same). I feel like there probably needs to be some code in this function to address that, or we'll have to have cleaned the database.

Have you tested this sync in somewhere with a full copy of production? @scvannost

cidc_api/models/templates/sync_schemas.py
```diff
@@ -0,0 +1,226 @@
...
+    for partic in participants:
+        # special handling for name change
+        partic["trial_participant_id"] = partic.pop("participant_id")
+
+        try:
+            # need to keep this for samples
+            new_partic = session.merge(
+                Participant(
+                    trial_id=new_trial.protocol_identifier,
+                    **{
```
jim-bo, line 195: Just a stylistic comment: I think I understand this argument expansion here but it might be helpful to comment this block of code explaining what it is doing, or perhaps even converting it to a function with a docstring for future explainability. I'll defer to you on the final call, not a deal breaker

cidc_api/models/templates/sync_schemas.py
```diff
@@ -0,0 +1,226 @@
...
+
+        samples = partic.get("samples", [])
+        for sample in samples:
+            try:
+                session.merge(
+                    Sample(
+                        trial_id=new_trial.protocol_identifier,
+                        cimac_participant_id=new_partic.cimac_participant_id,
+                        shipment_manifest_id=shipment_map[sample["cimac_id"]],
+                        **{
```
jim-bo, line 216: Same as above, it'd be good to comment what this expansion is doing

jim-bo (changes requested):
So in general I think this PR Is OK to merge given its light integration with the core system, I don't see a strong opportunity for this to have a negative impact if there is a mistake. However I would request that you test the full synchronization of the production database locally, before we merge and deploy the code.

cidc_api/models/templates/sync_schemas.py
scvannost: It will use the "last" name, which one will depend on the return order of the query
I could specify an `order by _created` to always use the most recent one

cidc_api/models/templates/sync_schemas.py
scvannost: [`session.merge`](https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session.merge) here will try to handle combining the entries using the primary keys, which are [trial_id and manifest_id](https://github.com/CIMAC-CIDC/cidc-api-gae/pull/549/files#diff-b1fbf246c0e7e54177b44319807161904f88c2aa8d0f7d7b5dacab17946138a7R201-R204)

PR #567, opened 2021-09-01
CIDC-1067 Add admin controls for relational Clinical Trials; add relational __all__'s

scvannost:
## What

Extends CIDC-1067 to allow for relational `ClinicalTrial`s to be edited along with the `TrialMetadata`s from the admin panel

## Why

For testing of new relational things, need to be able to add `Cohort`s and `CollectionEvent`s to then reference in the manifests. Currently, the only way to do this is to add them via the admin panel to the JSON blobs and then manually trigger a sync to propagate those changes into the relational bounds -- even then there is no way to remove any allowed cohort / collection event names.

## How

* `models.tempates.utils.remove_record_batch()` to remove arbitrary entries and an overarching `models.templates.utils.in_single_transaction()` to handle all updates in a single transaction to be rolled back if any errors occur
* compare old entries to the updated JSON returned from UI manually for cohorts / collection events
* add relational checks to the `/resources/trial_metadata` `POST`/`PATCH` tests
* fix return format from `/admin/load_from_blobs`
* change `ClinicalTrial.allowed_<...>_names` properties to not rely on the `relationship` and its lazy loading when not within a session
* also added `__all__`s to all `models.templates` files

## Remarks

There is no current direct testing of `insert_record_batch` so no test was written for `remove_record_batch`.  
Only a smoketest for correct ordering and a single commit was written for `in_single_transaction`.  
Finally, via the update trial endpoint testing, the tests do pass through these functions, providing bare minimal testing.

## Checklist

Please include and complete the following checklist. You can mark an item as complete with the `- [x]` prefix:

- [ ] Tests - Added unit tests for new code, regression tests for bugs and updated the integration tests if required
- [x] Formatting & Linting - `black` and `flake8` have been used to ensure styling guidelines are met
- [x] Type Annotations - All new code has been type annotated in the function signatures using type hints
- [x] Docstrings - Docstrings have been provided for functions
- [ ] Documentation - README has been updated to explain major changes & new features
- [x] Package version - Manually bumped the API package version in [setup.py](https://github.com/CIMAC-CIDC/cidc-api-gae/blob/master/setup.py#L21)


roshni-b (approved):
looks good!

PR #632, opened 2021-12-07
CIDC-1187 1198 replace conditional IAM logic with ACL logic for downloads only

scvannost:
makes `GOOGLE_DATA_BUCKET` stale except for some hanging old migrations

## What

- `remove` all logic associated with download logic ie conditional IAM permissions
- `add` ACL logic for downloads instead using `object.acl`'s `grant_reader` and `revoke_reader` for any `object : [gcs.Blob, gcs.Bucket]`
- `remove` all lister permission as no longer needed with ACL instead of IAM
- `add` admin endpoint to call already existing function to grant all download permissions

## Why

[CIDC-1187](https://dfcijira.dfci.harvard.edu:8443/browse/CIDC-1187) and [CIDC-1198](https://dfcijira.dfci.harvard.edu:8443/browse/CIDC-1198)

## How

In place change of the API layer between the Permissions table and the calls to GCS to implement said permissions.
Leave all existing expiring IAM code as it is used for upload and intake access functions -- conversely, remove all non-expiring and conditional IAM code as it was only used in the download access functions.
Instead of calls to expiring IAM code, change download access functions to directly call changes on the `acl` attribute of each blob that the permission applies to.
Manual testing showed that object lister permission is not required on the bucket when using an ACL-based permissions system, so remove all code related to handling lister permissions.

## Remarks

As the bucket itself is not stored in the GCS URI data in the database per the schemas, no migration or modification of the existing database is needed.
Only truly "new" code is the admin endpoint, which undergoes general endpoint testing -- the function it calls is smoke tested already to correctly call `gcloud_client.grant_download_access`.
All other modified code also have its corresponding test updated for the new internal functionality -- mocking ACL calls instead of IAM calls for all download access testing.

## Checklist

Please include and complete the following checklist. You can mark an item as complete with the `- [x]` prefix:

- [x] Tests - Added unit tests for new code, regression tests for bugs and updated the integration tests if required
- [x] Formatting & Linting - `black` and `flake8` have been used to ensure styling guidelines are met
- [x] Type Annotations - All new code has been type annotated in the function signatures using type hints
- [x] Docstrings - Docstrings have been provided for functions
- [ ] Documentation - README has been updated to explain major changes & new features
- [x] Package version - Manually bumped the API package version in [setup.py](https://github.com/CIMAC-CIDC/cidc-api-gae/blob/master/setup.py#L21)


scvannost:
currently failing due to -0.4% decrease in test coverage despite maintainability rising from B to A
this is likely because the new admin endpoint has no direct test

roshni-b (approved):
Looks good to me!

